PR #95, opened 2017-03-04
[WIP] Size Inference for Appender

sppalkia:
Infers the size of the Appender so memory does not need to be reallocated.

Not yet finished.


sppalkia:
Will reopen this after it's implemented.

Issue #109, opened 2017-03-16
Add setup.py for both weld and grizzly Python packages to make them installable


wesm:

deepakn94:
Addressed by #112.

PR #112, opened 2017-03-17
Better Python packaging

deepakn94:
@wesm, can you take a quick look? Thanks!

### DIFF
--- a/docs/python.md
+++ b/docs/python.md
@@ -2,10 +2,16 @@
 
 The Python API is found under the `python/` directory. It provides convinient wrapper objects for the low-level C API, as well as utilities for building Weld computations and composing Python libraries.
 
-To use the Python API, set the `PYTHONPATH` environment variable so it can find the `weld` Python package:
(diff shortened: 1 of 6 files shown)


deepakn94:
(looks to address #109)

jremmons:
@deepakn94 you should probably check to make sure that `WELD_HOME` environment variable is set in the `__init__` functions (python/grizzly/encoders.py for example). 

Also, I don't see a way to globally install weld on the user's systems in the docs (perhaps this can be done manually though)? Might be nice so that the user can install the python bindings without having to worry about other environment variables. 

Otherwise, I think this will do what you want. 

One more thing, setuptools will copy the contents of `python/weld` and `python/grizzly` to the `site-packages` dir of user's python installation when you run `python setup.py install`. If you guys update the python code, the users will need to git pull, then rerun `python setup.py install` to see the changes. Might want to make a note of this in the readme? Up to you though.

deepakn94:
Thanks @jremmons!

PR #114, opened 2017-03-18
Minor fixing of current Weld Python bindings

viirya:
The tutorial document has a minor issue which refers to incorrect path of Weld Python bindings. This is going to fix it.


sppalkia:
Hi @viirya,

Thanks for pointing this out! I committed a slightly modified version of this patch (instead of using `WELD_HOME`, we set `PYTHONPATH`, and I updated the doc in a few other places as well).

sppalkia:
This has been fixed; thanks for pointing it out!

PR #115, opened 2017-03-20
Fix build.rs not exiting if make commands fail

mateiz:
Before this, it just silently swallowed the errors in make.


deepakn94:
Thanks @mateiz, merging this.

PR #119, opened 2017-03-21
Use WELD_HOME for finding libraries

sppalkia:
This changes the build process to require that the `libweld` and `libweldrt` libraries live under `$WELD_HOME/target/debug` and `$WELD_HOME/target/release` respectively. The purpose of this is that `LD_LIBRARY_PATH` seems to require changing the system integrity protection on macOS for use with Python, which is undesirable.

@mateiz @deepakn94

### DIFF
--- a/python/weld/bindings.py
+++ b/python/weld/bindings.py
@@ -4,6 +4,7 @@
 
 from ctypes import *
 
+import os
 import platform
 import copy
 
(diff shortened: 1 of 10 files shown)


python/weld/bindings.py
```diff
@@ -17,24 +18,27 @@
...
+if home[-1] != "/":
+    home += "/"
+path = home + "target/debug/" + path
+
 # Load the Weld Dynamic Library.
 weld = CDLL(path)
 
 # Used for some type checking carried out by ctypes
 
-
```
deepakn94: I think Flake8 requires these line breaks

deepakn94:
Looks good (pulled this branch and everything works on my machine).

Can you fix `docs/python.md` and `python/README.md` as well to reflect the fact that you no longer need to set the `LD_LIBRARY_PATH` for Python applications to work?

deepakn94:
You can also remove the `export LD_LIBRARY_PATH=...` from .travis.yml if it's no longer needed.

rahulpalamuttam:
@sppalkia tests pass on linux.

PR #120, opened 2017-03-23
Python encoder

deepakn94:
@sppalkia: just as an FYI, not ready to merge.

Major issues right now:
- Some benchmarks don't yet work correctly
- Encoder / decoder much slower than C++ encoder / decoder


wesm:
Why would you do serialization / deserialization in Python when you can do it in C/C++? Do you have to copy the memory into a new memory space, or is zero-copy handoff possible?

wesm:
You might consider using https://github.com/spacetelescope/asv for performance tracking -- this is what we use in pandas for microbenchmarking and catching performance regressions.

deepakn94:
The main problem is that we need these two header files:

```#include "Python.h"
#include "numpy/arrayobject.h"```

It seemed like finding these header files is highly dependent on how the user's Python has been installed (for example, the user could use be using `virtualenv`), and we thought having everything in Python can help us route around this problem.

**Comment by wesm** [comment]
I don't think you're going to be able to avoid Python C extensions and linking to NumPy for very long -- best to sort out the build toolchain problems up front.

* For Python.h, it depends on your build system. If you are using distutils, this is handled automatically
* For numpy/arrayobject.h, you use `numpy.get_include`

On my system, I have

```
In [2]: numpy.get_include()
Out[2]: '/home/wesm/anaconda3/envs/py27/lib/python2.7/site-packages/numpy/core/include'
```

wesm:
In pyarrow, we are using CMake modules to find these dependencies 

https://github.com/apache/arrow/tree/master/python/cmake_modules

wesm:
FWIW, your primary audience for Weld in the short term are much more likely to be conda users than pip/virtualenv users -- I think you're going to want to figure out `conda install weld` sooner than `pip install weld` (which probably needs to involve manylinux1 wheels on Linux)

deepakn94:
Ok, great, I will look into this. Thanks for all the help!

deepakn94:
Closing.

PR #122, opened 2017-03-26
fix the decode bug in NumpyArrayDecoder

hustnn:
get real data with "ptr" field of WeldVec


deepakn94:
Thanks @hustnn, merging this in.

PR #129, opened 2017-04-07
Use sudo to create a symlink in /usr/local/bin

dobachi:
In case of my environment(Ubuntu14.04LTS), "sudo" was required to create a symlink in /usr/local/bin.


deepakn94:
Thanks @dobachi, I added a note to the main README saying "sudo" might be needed at the start of the command!

PR #130, opened 2017-04-10
Minor fix for examples of appender and merger

dobachi:
I think that the argument of second "merge" should be "b2" in examples of appender and merger.
(I understood that the purpose of these examples are to show the usage of each expression step by step.)


sppalkia:
Ah, good catch. Thanks for the fix! Merging this in.

PR #131, opened 2017-04-20
Add support for LLVM 4.0 (still supports 3.8+)

cirla:
- Use `LLVMLinkModules2` added in 3.8 (currently supported version) in place of deprecated `LLVMLinkModules`.
- Get LLVM version from `llvm-config` instead of hardcoding.
- Update README
- Update .travis.yml to:
  - Install binary versions of `numpy` and `pandas` to speed up Travis CI builds
  - Test against both LLVM 3.8 and 4.0

### DIFF
--- a/weld_rt/cpp/Makefile
+++ b/weld_rt/cpp/Makefile
@@ -1,5 +1,5 @@
 OS = $(shell uname -s)
-LLVM_VERSION = 3.8
+LLVM_VERSION = $(shell llvm-config --version | cut -d . -f 1,2)
 
 CFLAGS_RT = -O3 -Wall -fno-use-cxa-atexit -fPIC
 CFLAGS_BUILDER = ${CFLAGS_RT} -flto
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,19 +8,22 @@ python:
 addons:
   apt:
     sources:
+      - sourceline: 'deb http://llvm.org/apt/trusty/ llvm-toolchain-trusty-3.8 main'
+        key_url: 'http://llvm.org/apt/llvm-snapshot.gpg.key'
       - llvm-toolchain-trusty-4.0
     packages:
       - gcc
+      - make
+      - llvm-3.8
+      - llvm-3.8-dev
+      - clang-3.8
       - llvm-4.0
       - llvm-4.0-dev
       - clang-4.0
-      - make
       - lib32z1-dev
 
 install:
-  - sudo rm -rf /usr/bin/llvm-config
-  - sudo rm -rf `which llvm-config`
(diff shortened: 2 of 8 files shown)


sppalkia:
Hi @cirla,

Thanks so much for the contribution! Let us know if you need any kind of support from our end (questions about the codebase, integration tests, etc.).

Seems like the Travis CI is failing because it doesn't have LLVM 4.0 installed; could you get this patch to work if one or the other LLVM versions is not installed? Once that works, we can extend the CI instance to test whether these are missing/ensure the tests work on both versions.

Thanks again!

cirla:
Sure! I'll update the PR to pass Jenkins build.

cirla:
It's a pain to get Travis CI to run against both versions as the `llvm_sys` crate version is tied to the LLVM version. I've verified manually that builds and tests pass with LLVM 3.8/llvm_sys 38.0.1 and updated Travis to run against LLVM 4.0/llvm_sys 40.0.0.

weld_rt/cpp/Makefile
```diff
@@ -1,5 +1,5 @@
 OS = $(shell uname -s)
-LLVM_VERSION = 3.8
+LLVM_VERSION = $(shell llvm-config --version | cut -d . -f 1,2)
```
sppalkia: This doesn't seem to work for me on MacOS with `brew install llvm`; the executables produced by this install don't have a `4.0` prefix.

sppalkia:
Hi @cirla,

After some discussion, I think this is the right thing to do for now:

* Keep LLVM 3.8 as the default when compiling Weld
* Merge this in with documentation for how to support LLVM 4.0 if people want to use it (but by default, everything works with LLVM 3.8 "out of the box"). 

If you make those changes, I can go ahead and merge this in. Thanks!

cirla:
Thanks for the feedback! As it is, it should now work out of the box with LLVM 3.8 (I've updated the README appropriately). Also, the Travis CI builds now test both 3.8 and 4.0 (see [most recent passing build](https://travis-ci.org/weld-project/weld/builds/226626293)).

weld_rt/cpp/Makefile
cirla: Hmm this shouldn't change behavior if 3.8 is the installed version as the script as it is already appends `-3.8`. This just allows it to also work with other versions, so long as `llvm-config` is on the `PATH`. On my MacOS, installing via `brew install llvm` creates  `/usr/local/opt/llvm/bin/clang-4.0`.

sppalkia:
This looks great! I'll merge it in. Thanks for the contribution, @cirla!

PR #132, opened 2017-04-20
[WIP] Support Python 3 in addition to Python 2

cirla:
# Work in Progress; Do Not Merge

Add support to work in both Python 2 and Python 3

**TODO:**
 - Get Travis CI to run for multiple python versions. May have to change `language:` to `python` and install rust during the `install:` phase since testing is only done against latest rust anyway.
 - String operations (e.g. `WeldSeries.str.slice`) are still wonky and not working quite right in Python 3 with unicode strings. Differing numpy dtypes between Python 2.7 (`'S'`) and 3.6 (`'U'`) when loading CSVs. Need to determine work to be done to either:
    1.  Fully support unicode in both python 2.7 and 3.6 (perhaps as `UCS4`/`WeldInt`/`i32`), but how do we know when decoding if it's an array of ints or an array of UCS4 codepoints?
    2. Encode unicode as UTF-8 bytestrings between numpy and weld, though this will still cause string operations such as `slice` to function incorrectly on strings containing non-ASCII characters (e.g. chopping a 2-byte codepoint in half).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,8 @@ sudo: required
 dist: trusty
 
 python:
-  - 2.7
+  - "2.7"
+  - "3.6"
 
 addons:
   apt:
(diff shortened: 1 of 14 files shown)


deepakn94:
Hi @cirla, thank you for your contributions to Weld! Is this ready to be reviewed?

sppalkia:
@circla Closing this for now since it's somewhat outdated -- feel free to open it again once its updated :)

PR #137, opened 2017-04-30
Implement exists

rahulpalamuttam:
Given a dict and a key returns true if the key exists and false if it does not.


deepakn94:
Looks good; one minor nit: can you re-name "Exists" to something like "KeyExists" to make it easier to infer what this op does?

rahulpalamuttam:
@deepakn94 changed name to keyexists.

deepakn94:
Looks good, thanks. Merging in now.

Issue #134, opened 2017-04-25
Support arbitrary interleaving of horizontal and vertical fusion rules


deepakn94:
(Ideally we would be able to support arbitrary interleaving of all rules)

deepakn94:
Addressed by #139

PR #139, opened 2017-05-03
Apply transforms as part of optimization passes

deepakn94:


deepakn94:
PS: Do you know of a good way to specify these passes as part of some input configuration? Or is this not useful?

deepakn94:
(addresses #134)

sppalkia:
You could probably make a table of all the transforms, which maps a string name to each transform. That's similar to what clang does too (e.g., you can pass something like `-ftree-vectorize` to turn on the `tree-vectorize` transform, and `no-tree-vectorize` to turn the transform off).

mateiz:
This is great as a start but I'd consider creating a Pass class that has a name, etc. Then we can print nicer debugging info about passes or disable particular ones.

deepakn94:
Alright @matei and @sppalkia, tried to organize the code better.

deepakn94:
Also forgot to note that I plan on adding functionality that turns passes on and off in a separate PR...

deepakn94:
Actually just added the ability to specify optimization passes as part of the config. This works from Python too.

weld/passes.rs
```diff
@@ -0,0 +1,67 @@
...
+        }
+        Ok(())
+    }
+
+    pub fn pass_name(&self) -> String {
+        self.pass_name.clone()
+    }
+}
+
+pub fn get_pass(pass_name: String) -> WeldResult<Pass> {
```
mateiz: Why is pass_name a String instead of &str?

weld/passes.rs
```diff
@@ -0,0 +1,67 @@
...
+    match pass_name.as_ref() {
```
mateiz: Might be nicer to put these in a map with lazy_static.

weld/passes.rs
```diff
@@ -0,0 +1,67 @@
+use super::ast::*;
+use super::error::*;
+use super::transforms;
+
+
+pub struct Pass {
+    transform_fns: Vec<fn(&mut Expr<Type>)>,
+    transform_names: Vec<String>,
```
mateiz: This is kind of hard to maintain because the vectors have to be the same length; I'd create a separate "Transformation" type that has both a name and a function, or drop the names.

deepakn94:
@mateiz, addressed. Let me know if anything looks amiss.

weld/passes.rs
```diff
@@ -0,0 +1,58 @@
...
+    }
+}
+
+lazy_static! {
+    pub static ref OPTIMIZATION_PASSES: HashMap<&'static str, Pass> = {
+        let mut m = HashMap::new();
+        m.insert("inline-apply", Pass {
+            transform_fns: vec![transforms::inline_apply],
+            pass_name: String::from("inline-apply")
+        });
```
mateiz: Nit: I'd consider cleaning these up by creating a `Pass::new` constructor. For example, You could then do `m.insert("inline-apply", Pass::new("inline_apply, vec![inline_apply]))`.

weld/passes.rs
```diff
@@ -0,0 +1,58 @@
+use super::ast::*;
+use super::error::*;
+use super::transforms;
+
+use std::collections::HashMap;
+
+
+pub struct Pass {
+    transform_fns: Vec<fn(&mut Expr<Type>)>,
```
mateiz: Could just name this "transforms"

weld/conf.rs
```diff
@@ -27,3 +30,12 @@ pub fn parse_memory_limit(s: CString) -> i64 {
         Err(_) => DEFAULT_MEMORY_LIMIT,
     }
 }
+
+/// Parses the optimization passes. Returns the default if the string cannot be parsed.
+pub fn parse_optimization_passes(s: CString) -> Vec<String> {
+    let s = s.into_string().unwrap();
+    match s.as_ref() {
+        "" => DEFAULT_OPTIMIZATION_PASSES.to_vec().iter().map(|e| e.to_string()).collect(),
```
mateiz: Maybe DEFAULT_OPTIMIZATION_PASSES should be a static Vec<String> and we should just clone it. Not sure whether that's easier.

mateiz:
Looks OK overall, but I just made a few small comments on things that may make it clearer.

mateiz:
BTW feel free to just merge it once you've fixed those.

PR #140, opened 2017-05-05
Fix Code Generation for Negate on floats

sppalkia:
This fixes an issue where LLVM code generation was broken for the `Negate` operator on the `f32` and `f64` types.


jjthomas:
looks good, merging

PR #142, opened 2017-05-12
Fixes a Transform Bug in Loop Fusion

sppalkia:
Also adds better error reporting for missing symbols during `uniquify`

### DIFF
--- a/weld/ast.rs
+++ b/weld/ast.rs
@@ -444,8 +444,7 @@ impl<T: TypeBounds> Expr<T> {
                     if let Some(lv) = sym_map.get(l) {
                         Ok(**lv == *r)
                     } else {
-                        Err(WeldError::new("undefined symbol when comparing expressions"
-                            .to_string()))
+                        weld_err!("undefined symbol {} when comparing expressions", l)
                     }
                 }
                 _ => Ok(false), // all else fail.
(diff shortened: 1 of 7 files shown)


weld/ast.rs
```diff
@@ -523,21 +522,22 @@ impl<T: TypeBounds> Expr<T> {
 
     /// Recursively transforms an expression in place by running a function on it and optionally replacing it with another expression.
     pub fn transform_and_continue<F>(&mut self, func: &mut F)
-        where F: FnMut(&mut Expr<T>) -> Option<(Expr<T>, bool)>
+        where F: FnMut(&mut Expr<T>) -> (Option<Expr<T>>, bool)
```
mateiz: What's the motivation for this change? Is the bool about whether to continue recursing down or something?

weld/ast.rs
```diff
@@ -523,21 +522,22 @@ impl<T: TypeBounds> Expr<T> {
...
+            (Some(e), false) => {
                 *self = e;
             }
-            None => {
+            (None, true) => {
                 for c in self.children_mut() {
                     c.transform_and_continue(func);
                 }
             }
+            (None, false) => {}
```
mateiz: This whole switch statement is kind of weird now; why don't we have an if on the first part of the tuple (replace `*self` or not) and then a second if on the second part (transform children)?

weld/ast.rs
sppalkia: Yup, the `bool` states whether to recurse down and the expression is what to replace it with. Before, erroring out was hard since there was no way of saying "don't replace the expression and don't recurse down."

weld/ast.rs
sppalkia: Hm, the reason for this is that if you do end up replacing self, you want to call the transform on yourself (to match on the replaced expression); otherwise, you call the transform on each child.

mateiz: OK, that makes sense then.

sppalkia:
Cool, merged this.

deepakn94:
Hmm -- I think this commit broke some of the Pandas workloads. (seeing a segfault when running `python get_population_stats_simplified_grizzly.py`)

PR #143, opened 2017-05-12
Add decoding for 2D arrays of doubles in Grizzly

jialinding:


sppalkia:
Very cool, thanks! Merging this in.

PR #144, opened 2017-05-17
Annotations

deepakn94:


deepakn94:
@mateiz This is still work-in-progress, but feel free to leave comments!

mateiz:
Looks like you didn't `git add` all the files?

deepakn94:
Nope, I've `git add`ed everything I wanted to.

mateiz:
Oh weird, it was only showing 2 files before. Now it shows all of them.

weld/ast.rs
```diff
@@ -75,11 +75,136 @@ impl fmt::Display for ScalarKind {
...
+            num_keys: None,
+        }
+    }
+
+    pub fn get_builder_implementation(&self) -> Option<BuilderImplementationKind> {
+        self.builder_implementation.clone()
+    }
+
+    pub fn set_builder_implementation(&mut self,
+                                      builder_implementation: &str) {
```
mateiz: Would be better if this takes a BuilderImplementationKind instead of a string; why do the parsing here instead of in the parser?

mateiz: Also I wonder whether these should take in an option; is there any case for clearing / removing annotations? Maybe not...

weld/pretty_print.rs
```diff
@@ -36,8 +36,8 @@ impl PrintableType for Type {
                 res
             }
             Builder(Appender(ref t)) => format!("appender[{}]", t.print()),
-            Builder(DictMerger(ref kt, ref vt, op)) => {
-                format!("dictmerger[{},{},{}]", kt.print(), vt.print(), op)
+            Builder(DictMerger(ref kt, ref vt, op, ref annotations)) => {
+                format!("dictmerger[{},{},{}] #[{}]", kt.print(), vt.print(), op, annotations)
```
mateiz: Don't print the annotations unless there are some set IMO

mateiz:
How hard is it to make the syntax such that annotations appear before the type / statement / etc, not after? In most languages it's before.

Also I'd consider making the syntax more Java- and Python-like: `@(annotations)`.

deepakn94:
I think this is ready for review now.

Couple of points I would like feedback on:
- I'm not currently attaching annotations to their corresponding `Expr` objects -- should this be added before we merge in the first pass? (in particular, if you have an annotation on a builder, you could also have the same annotation on the corresponding `NewBuilder` object -- what makes most sense here?)
- Since `Builder` needs the `Eq` and `Hash` traits implemented, I had to make branch_selectivity an `Option<i32>` -- this is awkward however. What's the best way of dealing with this?

Any other feedback appreciated as well!

weld/ast.rs
```diff
@@ -75,6 +75,181 @@ impl fmt::Display for ScalarKind {
...
+        };
+        f.write_str(text)
+    }
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub struct Annotations {
+    builder_implementation: Option<BuilderImplementationKind>,
+
+    predicate: Option<bool>,
```
sppalkia: Predication is a transform, so maybe this shouldn't be an annotation?

weld/ast.rs
```diff
@@ -75,6 +75,181 @@ impl fmt::Display for ScalarKind {
...
+    vectorize: Option<bool>,
+    tile_size: Option<i32>,
+    grain_size: Option<i32>,
+
+    size: Option<i32>,
```
sppalkia: Probably good to make these `i64` since the sizes can get pretty big (assuming this is in bytes)

weld/ast.rs
```diff
@@ -75,6 +75,181 @@ impl fmt::Display for ScalarKind {
...
+    branch_selectivity: Option<i32>, // Fractions of 10,000.
```
sppalkia: Maybe get rid of `Eq` as a trait and just make this `f32`?

weld/ast.rs
```diff
@@ -75,6 +75,181 @@ impl fmt::Display for ScalarKind {
...
+            vectorize: None,
+            tile_size: None,
+            grain_size: None,
+            size: None,
+            branch_selectivity: None,
+            num_keys: None,
+        }
+    }
+
+    pub fn get_builder_implementation(&self) -> Option<BuilderImplementationKind> {
```
sppalkia: Rename `get_builder_implementation` -> `builder_implementation`

https://github.com/rust-lang/rfcs/blob/master/text/0344-conventions-galore.md#gettersetter-apis

Is there a reason why we return a clone of the annotation instead of a reference?

weld/ast.rs
```diff
@@ -75,6 +75,181 @@ impl fmt::Display for ScalarKind {
...
+        self.builder_implementation.clone()
+    }
+
+    pub fn set_builder_implementation(&mut self,
+                                      builder_implementation: BuilderImplementationKind) {
+        self.builder_implementation = Some(builder_implementation);
+    }
+
+    pub fn get_predicate(&self) -> Option<bool> {
```
sppalkia: See above.

weld/ast.rs
```diff
@@ -75,6 +75,181 @@ impl fmt::Display for ScalarKind {
...
+        self.predicate.clone()
+    }
+
+    pub fn set_predicate(&mut self, predicate: bool) {
+        self.predicate = Some(predicate)
+    }
+
+    pub fn get_vectorize(&self) -> Option<bool> {
```
sppalkia: See above.

weld/ast.rs
```diff
@@ -75,6 +75,181 @@ impl fmt::Display for ScalarKind {
...
+        self.vectorize.clone()
+    }
+
+    pub fn set_vectorize(&mut self, vectorize: bool) {
+        self.vectorize = Some(vectorize)
+    }
+
+    pub fn get_tile_size(&self) -> Option<i32> {
```
sppalkia: See above.

weld/ast.rs
```diff
@@ -75,6 +75,181 @@ impl fmt::Display for ScalarKind {
...
+        self.tile_size.clone()
+    }
+
+    pub fn set_tile_size(&mut self, tile_size: i32) {
+        self.tile_size = Some(tile_size)
+    }
+
+    pub fn get_grain_size(&self) -> Option<i32> {
```
sppalkia: See above.

weld/ast.rs
```diff
@@ -75,6 +75,181 @@ impl fmt::Display for ScalarKind {
...
+
+    pub fn get_selectivity(&self) -> Option<i32> {
+        self.branch_selectivity.clone()
+    }
+
+    pub fn set_selectivity(&mut self, selectivity: i32) {
+        self.branch_selectivity = Some(selectivity)
+    }
+
+    pub fn get_num_keys(&self) -> Option<i32> {
```
sppalkia: See above.

weld/ast.rs
```diff
@@ -75,6 +75,181 @@ impl fmt::Display for ScalarKind {
...
+
+    pub fn get_size(&self) -> Option<i32> {
+        self.size.clone()
+    }
+
+    pub fn set_size(&mut self, size: i32) {
+        self.size = Some(size)
+    }
+
+    pub fn get_selectivity(&self) -> Option<i32> {
```
sppalkia: See above.

weld/ast.rs
```diff
@@ -75,6 +75,181 @@ impl fmt::Display for ScalarKind {
...
+
+    pub fn get_grain_size(&self) -> Option<i32> {
+        self.grain_size.clone()
+    }
+
+    pub fn set_grain_size(&mut self, grain_size: i32) {
+        self.grain_size = Some(grain_size)
+    }
+
+    pub fn get_size(&self) -> Option<i32> {
```
sppalkia: See above.

weld/ast.rs
```diff
@@ -75,6 +75,181 @@ impl fmt::Display for ScalarKind {
...
+
+    pub fn get_num_keys(&self) -> Option<i32> {
+        self.num_keys.clone()
+    }
+
+    pub fn set_num_keys(&mut self, num_keys: i32) {
+        self.num_keys = Some(num_keys)
+    }
+
+    pub fn is_empty(&self) -> bool {
```
sppalkia: Is there a way to do this so the compiler throws an error if we add new annotations and forget to add them here (e.g., something like a `MatchError`?) Maybe it's not possible.

weld/ast.rs
```diff
@@ -75,6 +75,181 @@ impl fmt::Display for ScalarKind {
...
+            is_empty = false;
+        }
+        if let Some(_) = self.num_keys {
+            is_empty = false;
+        }
+        is_empty
+    }
+}
+
+impl fmt::Display for Annotations {
```
sppalkia: Same comment as for `is_empty`

weld/parser.rs
```diff
@@ -485,8 +487,115 @@ impl<'t> Parser<'t> {
         Ok(cast_expr)
     }
 
+    /// Parses annotations in the format "@(<annotation name>: <annotation value>,...)".
+    fn parse_annotations(&mut self, annotations: &mut Annotations) -> WeldResult<()> {
```
sppalkia: Can you add these to `language.md` describing the syntax, valid annotations, etc.

weld/ast.rs
deepakn94: I think the point of this is to specify whether you want to predicate a specific `if` expression. Similar with `vectorize`, etc.

weld/ast.rs
deepakn94: Good point, I'll fix this.

weld/ast.rs
deepakn94: Noted -- changed to returning a reference as well.

weld/ast.rs
deepakn94: Don't think so -- I'll add a comment in the `struct` definition though.

weld/ast.rs
deepakn94: Hmm yeah -- I don't think this is possible since `BuilderKind` needs the `Eq` and `Hash` traits implemented.

deepakn94:
@sppalkia, addressed your comments!

docs/language.md
```diff
@@ -48,6 +48,15 @@ Note that among the builders that take functions and or values, two types are id
 Making functions part of the type is slightly unusual because the functions passed in could be closures over some variables that exist at that point in the program.
 Right now the implementation doesn't allow this, and we have not tried to formalize what it means.
 
+In addition, it's possible to specify annotations on builder types: these could for example specify an implementation strategy for a builder. To specify an annotation on a `dictmerger`, one can use syntax like
```
sppalkia: It might be better to just have a subsection called Annotations and list the ones we support?

weld/ast.rs
sppalkia: Ah I see...and `Type` is used as a `HashMap` key in `llvm.rs`. I think eventually it might be good to just get rid of those traits for types but then we'd need a different way of storing a type in a map. This is okay for now.

sppalkia:
Also, make sure you resolve the conflicts and run `rustfmt` again.

PR #145, opened 2017-05-18
Add C UDF Support

sppalkia:
This PR adds support for calling C-style UDFs from within an Weld program. C UDFs are functions with with C-style symbol names. As long as the symbol name is linked with the executable, Weld can call into it.

* See `language.md` for a detailed discussion on the semantics of the UDF node.
* See `examples/cpp/udfs/` for an example of using UDFs in a C++ program.

TODO: Still need to figure out how to add integration tests for the UDFs; we somehow need to link a C-callable function into the integration test suite. Alternative designs for the UDF node are also welcome.

### DIFF
--- a/docs/language.md
+++ b/docs/language.md
@@ -186,7 +186,7 @@ macro filter(data, func) = (
 
 ## User Defined Functions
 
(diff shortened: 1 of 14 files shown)


jjthomas:
Took a quick look -- looks pretty good. You might want to add some info on where the C function needs to be located -- I think any function in the address space of the program that calls weld_module_compile/weld_module_run is fine, right? It can be statically/dynamically linked, etc.

jjthomas:
Also, I don't understand the challenge with the integration tests -- can you just pass a Rust function that has a C interface?

sppalkia:
@jjthomas yeah that's what I'll try.

sppalkia:
Okay @jjthomas, added a unit test.

jjthomas:
Looks good, feel free to merge.

Issue #110, opened 2017-03-16
Grizzly is Python 2.7 only


wesm:
It will be important to run on Python 3, preferably both 2.7 and 3.5/3.6 with a single codebase (the `six` module helps with this)

cirla:
I'm working on supporting 2.7 + 3.5/3.6 simultaneously in https://github.com/weld-project/weld/pull/132, though Unicode support muddies things a bit.

wesm:
While you're at it, it would be nice to plot a course to `conda install weld` and get all the Python things in a single `import weld` statement. This probably means a package structure like

```
weld/
    grizzy/ ...
```

cirla:
I'm also exploring what it will take to package the shared libs into the packages so binary wheels can be built and distributed.

wesm:
You can look at what we did in Apache Arrow with manylinux1: https://github.com/wesm/arrow/blob/master/python/manylinux1/build_arrow.sh

and https://github.com/wesm/arrow/blob/master/python/setup.py#L210

so all the shared libs (build with CMake) get bundled in the wheel. Probably possible to do something similar with OS X. We have an extra layer of complexity in that we also want to expose a C API for Arrow via the binary wheel (similar to the NumPy C API), but we're still working on that.

wesm:
conda is the easiest way since you can package `libweld` (the shared libraries) and `weld-python` (the Python package and C extensions) as separate components

wesm:
There seems to be some GitHub snafu right now so all the Apache git mirrors on GitHub are down at the moment

snakescott:
@cirla how is work on this going? I'm new to Weld, but Python 3.6 + packaging support lines up with my interests; is there a part of this work (and/or #132) that is sufficiently 

* well defined
* separable
* non-duplicative of current efforts

that I could try tackling it?

Thanks!

cirla:
@snakescott I haven't been actively working on this recently, but there are two separate things to tackle for #132:

1. Get Travis CI to run for multiple python versions.
   - May have to change `language:` in `.travis.yml` to `python` so that it spawns a separate job for each python version specified under the `python:` list, but then we'd need to install rust during the `install:` phase (i.e. `curl -sSf https://build.travis-ci.org/files/rustup-init.sh | sh -s -- --default-toolchain=$TRAVIS_RUST_VERSION -y`), but this would preclude us from testing against multiple versions of rust.
[...]

PR #148, opened 2017-05-31
Support different versions of python on Travis

snakescott:
Makes progress towards Python 3.6 support (#110). Once codebase is 3.6 compatible (e.g. print, unicode, etc), adding appropriate invocations to `script` in `.travis.yml` will be enough to build and test under both 2.7 and 3.6.

Once unicode & 3.6 support is working, better packaging of weld and grizzly and publishing artifacts to conda-forge would be nice.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,12 +2,10 @@ language: rust
 sudo: required
 dist: trusty
 
-python:
-  - 2.7
-
 addons:
   apt:
     sources:
+      - deadsnakes
       - sourceline: 'deb http://llvm.org/apt/trusty/ llvm-toolchain-trusty-3.8 main'
         key_url: 'http://llvm.org/apt/llvm-snapshot.gpg.key'
       - llvm-toolchain-trusty-4.0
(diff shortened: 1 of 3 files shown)


snakescott:
@cirla you are probably the right person to review this PR?

sppalkia:
Hey @snakescott,

Can you rebase this and make sure it passes Travis again (in particular, make sure you run `rustfmt` to autoformat the code). I did a quick scan through it and it looks good, but I'll read it more closely once that's done.

Thanks!
Shoumik

.travis.yml
```diff
@@ -21,16 +19,20 @@ addons:
...
+      - python2.7
+      - python3.6
 
 install:
-  - sudo rm -f `which llvm-config`
-  - export WELD_HOME=`pwd`
+  - pip install virtualenv==15.1.0
+  - mkdir -p .virtualenv
+  - export VENV_HOME=`pwd`/.virtualenv
+  - virtualenv $VENV_HOME/py2.7
```
cirla: It might be best to be explicit with both of these, e.g.:

```
- virtualenv $VENV_HOME/python2.7 --python=python2.7
- virtualenv $VENV_HOME/python3.6 --python=python3.6
```

You'll also want to `deactivate` below before `activate`-ing a different virtualenv. You can probably wrap it all in a helper script like:

```shell
 #!/bin/bash

$PYTHON_VERSIONS=$@

pip install virtualenv==15.1.0
mkdir -p .virtualenv
export VENV_HOME=`pwd`/.virtualenv

for v in $PYTHON_VERSIONS; do
  virtualenv "$VENV_HOME/python$v" --python="python$v"
  source "$VENV_HOME/python$v/bin/activate"
  pip install --upgrade pip setuptools wheel
  pip install --only-binary=numpy,pandas -r requirements.txt
  deactivate
done
```

and then call:

```
  - init_virtualenvs.sh 2.7 3.6
```

sppalkia:
This looks good to me! @snakescott is it still WIP?

snakescott:
yeah I think this should be g2g

Issue #149, opened 2017-06-02
Travis CI Doesn't enforce format


sppalkia:
We might want to make Travis CI enforce `rustfmt`; this will keep commits clean when `rustfmt` is run.

@mateiz @jjthomas @deepakn94 what do you think?

mateiz:
Yup, let's do it!

sppalkia:
Resolved in #150

PR #150, opened 2017-06-02
Add `rustfmt` to Travis

sppalkia:
See #149. We might choose to not merge this in :)


sppalkia:
FYI, this travis build fails because of the `rustfmt` check. We might want to add a commit which formats the whole codebase...it will make merging other outstanding PRs annoying though.

PR #155, opened 2017-06-11
[WIP] Global dictmerger

deepakn94:
Still need to make the updates atomic, but comments welcome!

### DIFF
--- a/tests/integration_tests.rs
+++ b/tests/integration_tests.rs
@@ -900,6 +900,60 @@ fn simple_for_dictmerger_loop() {
     unsafe { weld_value_free(ret_value) };
 }
 
+fn simple_for_global_dictmerger_loop() {
+    #[derive(Clone)]
+    #[allow(dead_code)]
+    struct Pair {
+        ele1: i32,
+        ele2: i32,
+    }
+
+    #[allow(dead_code)]
+    struct Args {
+        x: WeldVec<i32>,
+        y: WeldVec<i32>,
+    }
+
+    let code = "|x:vec[i32], y:vec[i32]| tovec(result(for(zip(x,y), @(impl:global) dictmerger[i32,i32,+], \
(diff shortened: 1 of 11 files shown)


tests/integration_tests.rs
```diff
@@ -900,6 +900,60 @@ fn simple_for_dictmerger_loop() {
     unsafe { weld_value_free(ret_value) };
 }
 
+fn simple_for_global_dictmerger_loop() {
```
jjthomas: Is this function just a copy of `simple_for_dictmerger_loop` with the `global` annotation? If so, we should try to share the code between the two tests.

weld/resources/dictmerger_global.ll
```diff
@@ -0,0 +1,67 @@
...
+  %structSize = ptrtoint %$NAME* %structSizePtr to i64
+  %rawPtr = call i8* @new_merger(i64 %structSize, i32 1)
+  %newDict = call %$NAME @$NAME.new(i64 %capacity)
+  %bldPtr = bitcast i8* %rawPtr to %$NAME*
+  store %$NAME %newDict, %$NAME* %bldPtr, align 1
+  ret %$NAME.bld %bldPtr
+}
+
+; Append a value into a builder, growing its space if needed.
+define %$NAME.bld @$NAME.bld.merge(%$NAME.bld %bldPtr, %$KV_STRUCT %keyValue, i32 %workerId) {
```
jjthomas: Can we share this code with the local `dictmerger`?

deepakn94: Are you suggesting refactoring the linear probing logic to some sort of `dictmerger_prelude`?

jjthomas: Right, the `dictmerger` struct is the same in both cases, right?

PR #157, opened 2017-06-14
Constructors for Expressions

sppalkia:
Constructor functions for creating (typed) expressions. These functions are meant to be used in transform rules, and they automatically assign types to expressions given their subexpressions. This will both make it easier to write transform rules and reduce bugs (e.g., bugs where we replaced the `ExprKind` but not the `Type` in a transform have been common).


deepakn94:
Looks good to me.

PR #159, opened 2017-06-20
Vectorization

sppalkia:
Implements simple vectorization. This PR has a number of limitations which will be addressed shortly.

The gist of how it works: The vectorizer checks to make sure all the requirements for vectorization are met. If so, it changes the types of compatible expressions to the new `Vectorized` type.

Iterators in loops now have a kind: `VectorIter`, `ScalarIter`, or `FringeIter`. The `ScalarIter` indicates no vectorization. The `VectorIter` loads elements according to the usual iterator rules until it can't fill a vector because the iterator ends. The `FringeIter` runs scalar code which handles the "fringe" of the vector iterator.

Code generation is responsible for implementing the above iterators correctly.

For a detailed description of requirements and current limitations, see this doc:
https://github.com/sppalkia/weld/blob/vectors-2/docs/internals/vectorization.md


docs/internals/vectorization.md
```diff
@@ -0,0 +1,72 @@
+## Vectorization in Weld
```
mateiz: This is great, thanks for writing an overview doc

weld/partial_types.rs
```diff
@@ -10,6 +10,7 @@ use super::error::*;
 pub enum PartialType {
     Unknown,
     Scalar(ScalarKind),
+    Vectorized(ScalarKind),
     Vector(Box<PartialType>),
```
mateiz: As a nit, I think this Vectorized vs Vector is kind of confusing. Not sure how to change it except to change Vector to List or Array. Maybe we can make that change later so we don't delay this PR, but think about alternative names here.

weld/pretty_print.rs
```diff
@@ -26,6 +26,12 @@ impl PrintableType for Type {
             Scalar(I64) => "i64".to_string(),
             Scalar(F32) => "f32".to_string(),
             Scalar(F64) => "f64".to_string(),
+            Vectorized(Bool) => "<? x bool>".to_string(),
+            Vectorized(I8) => "<? x i8>".to_string(),
+            Vectorized(I32) => "<? x i32>".to_string(),
+            Vectorized(I64) => "<? x i64>".to_string(),
+            Vectorized(F32) => "<? x f32>".to_string(),
+            Vectorized(F64) => "<? x f64>".to_string(),
```
mateiz: Do we want to allow parsing these too? I'd consider allowing that and choosing a simpler syntax, e.g. `simd[i32]`.

Also on a separate node you can factor this code to implement pretty_print for `ScalarKind` only once (perhaps just implement `Display` on it).

weld/partial_types.rs
mateiz: Maybe one idea is to switch `Vectorized` to `Simd`. Or we can change `Vectorized` to `Vector` and `Vector` to `Seq` or `List` or something.

tests/integration_tests.rs
```diff
@@ -622,6 +622,32 @@ fn complex_parallel_for_appender_loop() {
     unsafe { weld_value_free(ret_value) };
 }
 
+fn simple_for_vectorizable_loop() {
```
mateiz: Do you actually check that this vectorized the loop?

This is another reason it may be good to implement parsing -- so we can have tests that directly execute some vectorized code.

mateiz: Also we should have unit tests for the other builders that support vectorization.

weld/partial_types.rs
sppalkia: I renamed this to `Simd`.

weld/pretty_print.rs
sppalkia: Added parsing. This may need to change in the future but it works for now (namely, since the length is not encoded in the AST, there is no way to specify literals like `<1, 2, 3, 4>`. I added a `broadcast` node so vectors composed of a single constant can be expressed (e.g,. `<1, 1, 1, 1>` can be expressed with `broadcast(1)`).

tests/integration_tests.rs
sppalkia: I've only implemented `merger` so far but `appender` is coming soon. Right now I'll just print something if it vectorizes (so I know all the merger loops are being vectorized now), but yes, checking the string is better so we don't need to print. I'll add that.

sppalkia:
@mateiz I addressed most of your comments (still need to add some tests and check whether things are getting vectorized by comparing their strings). I also added a few more features:

* Predication
* A broadcast node to represent vectors composed of a single constant
* Support for vectorizing functions which have more than one iterator (i.e., a `zip`)
* Parsing and type inference for vectors.

Also renamed `Vectorized` to `Simd`, and vector types now print as `simd[T]`.

weld/exprs.rs
```diff
@@ -217,6 +226,28 @@ pub fn if_expr(cond: Expr<Type>,
              ty)
 }
 
+pub fn select_expr(cond: Expr<Type>,
+               on_true: Expr<Type>,
+               on_false: Expr<Type>)
+               -> WeldResult<Expr<Type>> {
+    let err = weld_err!("Internal error: Mismatched types in select_expr");
+    if cond.ty != Scalar(ScalarKind::Bool) && cond.ty != Simd(ScalarKind::Bool) {
```
jjthomas: I see, you allow the input code to have `simd` expressions

tests/integration_tests.rs
```diff
@@ -622,6 +622,58 @@ fn complex_parallel_for_appender_loop() {
...
+    #[allow(dead_code)]
+    struct Args {
+        x: WeldVec<i32>,
+    }
+
+    let code = "|x:vec[i32]| result(for(x, merger[i32,+], |b,i,e| let a = 1; let a2 = a + 1; merge(b, e+a2)))";
+    let conf = default_conf();
+
+    let size = 1000;
```
jjthomas: Can you make the sizes of this and other vectorization tests not multiples of 4 so that fringing is tested?

weld/vectorizer.rs
```diff
@@ -0,0 +1,362 @@
...
+            }
+        }
+        if iter.kind != IterKind::ScalarIter {
+            return false;
+        }
+    }
+    true
+}
+
+/// Vectorizes the expression by changing it's type if the expression is a scalar.
```
jjthomas: it's -> its

tests/integration_tests.rs
```diff
@@ -649,6 +701,39 @@ fn simple_for_merger_loop() {
...
+    let y_data = vec![1; size as usize];
+
+    let ref input_data = Args {
+        x: WeldVec {
+            data: x_data.as_ptr(),
+            len: x_data.len() as i64,
+        },
+        y: WeldVec {
+            data: x_data.as_ptr(),
+            len: x_data.len() as i64,
```
mateiz: shouldn't this be y_data? I get an "unused variable" on y_data, although it is identical to x_data in this case

tests/integration_tests.rs
mateiz: Probably should try both a fringing and non-fringing test case

weld/vectorizer.rs
```diff
@@ -0,0 +1,362 @@
...
+    if new_expr.is_some() {
+        *e = new_expr.unwrap();
+    }
+
+    Ok(cont)
+}
+
+
+/// Checks basic vectorizability for a loop - this is a strong check which ensure that the only
+/// expressions which appear in a function body are arithmetic, identifiers, literals,and Let
```
mateiz: typo here?

docs/internals/vectorization.md
```diff
@@ -0,0 +1,74 @@
...
+* `broadcast(s)` broadcasts the scalar value in `s` to a vector.
+* `select(cond, on_true, on_false)` evaluates `cond`, `on_true`, and `on_false` and returns a value
+  based on whether `cond` is `true` or `false`.
+* The `simd[T]` type is used to specify SIMD vectors.
+
+### Current Limitations and To Dos
+
+* Only the `merger` builder is supported
+* Scatters and gathers are not supported (iterators must look at all elements), and index
+  computations are disallowed in the for loop body)
```
mateiz: It looks like the code actually allows these when it tries to vectorize, but then fails during LLVM compilation. E.g. try

```
|x:vec[i64]| result(for(x, merger[i64,+], |b,i,e| merge(b, e+i)))
```

weld/sir.rs
```diff
@@ -355,11 +374,12 @@ impl fmt::Display for ParallelForIter {
...
             write!(f,
-                   "iter({}, {}, {}, {})",
+                   "iter({}, {}, {}, {}, {})",
                    self.data,
                    self.start.clone().unwrap(),
                    self.end.clone().unwrap(),
-                   self.stride.clone().unwrap())
+                   self.stride.clone().unwrap(),
+                   self.kind)
         } else {
```
mateiz: Nit: add another case for when self.kind is set so that we see fringe vs vectorized in the SIR programs

weld/sir.rs
```diff
@@ -441,6 +461,9 @@ fn sir_param_correction_helper(prog: &mut SirProgram,
                 Negate { ref child, .. } => {
                     vars.push(child.clone());
                 }
+                Broadcast { ref child, .. } => {
```
mateiz: document what this does here

weld/llvm.rs
```diff
@@ -387,16 +421,60 @@ impl LlvmGenerator {
...
+                            let arr_len = ctx.var_ids.next();
+                            let offset = ctx.var_ids.next();
+                            let final_idx = ctx.var_ids.next();
+
+                            ctx.code
+                                .add(format!("{} = call i64 {}.size({} {})",
+                                arr_len,
+                                data_prefix,
+                                &data_ty_str,
+                                data_str));
```
sppalkia: These are all because of `rustfmt`...should I change them? We've stopped using that tool because it changes how it formats code every minor release.

weld/sir.rs
sppalkia: I put a note on what this does in `ast.rs`; putting in `sir.rs` seems kind of weird.

sppalkia:
@jjthomas @mateiz addressed comments. @mateiz, the weird formatting is from `rustfmt`, which we've stopped using now (because the way it formats code changes every time they update the tool, so it makes commits very long). Maybe we can fix those things using the auto-formatter when it's stable?

mateiz:
We should just not use rustfmt and fix them by hand until it's good. It makes the code really hard to read.

sppalkia:
That works - I can do that in another PR since the entire codebase has the weird "ctx.code\n.add" thing - don't want to increase the diff for this file/make it inconsistent.

mateiz:
Looks like this has a merge conflict now

PR #160, opened 2017-06-20
Initial Arrow compatibility via pyarrow/Grizzly plus build cleanup

snakescott:
* Add cython extension, encoder, and decoder for one dimension Arrow arrays 
* Update SeriesWeld to support pluggable encoder/decoder instead of being hardcoded to numpy
* Port grizzly tests to cover Arrow encoding/decoding as well

Since Arrow interop required a Cython extension also
* Build numpy_weld_converter via python extension instead of build.rs / Makefile
* Reorder travis build steps to build Rust code first
* Eliminate use of WELD_HOME in python code

Future work might include support for two dimensional Arrow objects, etc


snakescott:
@sppalkia

python/weld/arrow_compat.pyx
```diff
@@ -0,0 +1,44 @@
+from pyarrow.lib cimport Array
+from libcpp.memory cimport shared_ptr, make_shared
+from libc.stdint cimport uint8_t, int64_t, uintptr_t
+from ctypes import cast, addressof
+
+from pyarrow.lib cimport pyarrow_wrap_array, DataType
+from pyarrow.includes.libarrow cimport *
```
wesm: This is cool, I'm glad you got this working. We should think about some affordances in the public pyarrow Cython API to make this easier for you. I will have a closer look at the patch when I can

snakescott: Feedback definitely welcome, I'm certainly not a Cython expert!

sppalkia:
@circla Closing this for now since it's somewhat outdated -- feel free to open it again once its updated :)

Issue #163, opened 2017-07-11
Cannot reuse a weld input vector.


hvanhovell:
I am trying to iterate twice over the same vector, however this fails with a `BadIteratorLength` error. It seems that some of the state of the vector is mutated. The following code current fails:
```
|x:i32, ys:vec[i64]|
let a = result(for(ys, appender[i32], |b, i, y| merge(b, x)));
let b = result(for(ys, appender[i64], |b, i, y| merge(b, y)));
{a, b}
```

sppalkia:
Hmm, I'm not seeing this issue. I added some example code to the repository in #164 if you'd like to take a look at that; make sure a correct length is set in the vector struct.

hvanhovell:
@sppalkia thanks for looking into this. I am calling weld from java, and something might be getting lost in translation. I also just added a similar test case my local weld repo, and that seems to work properly. Lemme take another look at my code.

hvanhovell:
This issue is caused by padding. I was packing `x` (4 bytes) and the `ys`(16 bytes) into a 20 byte struct. Weld - on my platform at least - uses 8 byte alignment and pads the `x` value with an additional 4 bytes. It might be - slightly - useful to put this in the documentation.

Issue #162, opened 2017-07-11
Cannot materialize a struct of vector builders into a struct of vectors.


hvanhovell:
The following code tries to generate a struct of vectors:
```
|x:i32, ys:vec[i64]|
result(for(ys, {appender[i32], appender[i64]}, |bs, i, y| {merge(bs.$0, x), merge(bs.$1, y)}))
```
However it fails with the following error:
```
Error during LLVM code gen:
Non builder type {appender[i32],appender[i64]} found in GetResult
```
It seems that I cannot materialize a struct of vector builders into a struct of vectors.

hvanhovell:
cc @mateiz

sppalkia:
See #164 - I updated the doc slightly. For now, please call `result` on each struct element individually to get around this; we'll eventually add support for calling `result` on a struct of builders as well.

hvanhovell:
@sppalkia thanks for the quick reaction!

Issue #161, opened 2017-07-11
Composite builder example is broken


hvanhovell:
When I execute the composite builder example in the repl:
```
let b1 = appender[i32];
let b2 = appender[i32];
let data = [1, 2, 3];
let bs = for(
  data,
  {b1, b2},
  |bs: {appender[i32], appender[i32]}, i: i64, n: i32| {merge(bs.$0, n), merge(bs.$1, 2 * n)}
);
result(bs)
```
It fails with the following error:
```
thread 'main' has overflowed its stack
fatal runtime error: stack overflow
Abort trap: 6
```
Also note that the example in the documentation is not correct.

I am on mac OS X 10.11.6.

sppalkia:
Hi @hvanhovell,

Thanks for trying out the project and finding the issues! Some of the syntax errors are because the documentation was ported over from an old implementation of the project.

I fixed the stack overflow error in #164, along with the documentation example. Please don't hesitate to continue raising any more bugs you may encounter; we'll be sure to take a look at them.

-Shoumik

PR #164, opened 2017-07-11
Fixes from Issues

sppalkia:
* Fixes issue in #161 and update `langauge.md` with correct syntax.
* Addresses issue in #162 by disallowing `Result` on structs; call `Result` on each element of the struct instead (see updated `language.md` for example).
* Add C++ examples for how to retrieve and pass composite structures from/to Weld (using code from #163)

### DIFF
--- a/docs/language.md
+++ b/docs/language.md
@@ -123,17 +123,17 @@ result(b2)    # returns [2, 4, 6]
 ```
 
 ```
-# for expression with composite builder
-let b0 = appender[i32];
+# for with composite builder
 let b1 = appender[i32];
+let b2 = appender[i32];
 let data = [1, 2, 3];
 let bs = for(
   data,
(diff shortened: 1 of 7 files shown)


jjthomas:
Looks good, could you briefly explain what the stack overflow issue was?

sppalkia:
It was an issue with the uniquifier; basically, if you had an expression like this:

`let a = for([1], builder, |b, i, a| merge(b, a)); a` where `a` is redefined in a lambda, the stack of symbol numbers was being popped after leaving the scope of the lambda. So the `a` in the let statement would be assigned the same name.

The solution is to always use the current max ID assigned for new names so this does happen (i.e., popping the stack of symbol numbers makes sure we use the current symbol, and the max ID makes sure that names are unique).

PR #166, opened 2017-07-17
Add a new config flag for detailed logging

mateiz:
This also updates the REPL to use this flag when compiling instead, preventing lots of replication between llvm::compile_program and the REPL.


mateiz:
You could also just pass the LogLevel into them as part of Pass. We can add that later if you'd like.

mateiz:
BTW in reality I'd like to change this to a Logger that also prints the time and additional info.

PR #167, opened 2017-07-18
Add a function to determine the memory usage of a weld_value's run

mateiz:
This can be useful for debugging to understand how much memory some code allocated. One awkward thing is that we may free memory while a run is executing in the future, so this needs to be documented carefully so that users notice if it changes.

### DIFF
--- a/easy_ll/src/lib.rs
+++ b/easy_ll/src/lib.rs
@@ -60,8 +60,8 @@ impl From<NulError> for LlvmError {
     }
 }
 
-/// The type of our "run" function pointer.
-type RunFunc = extern "C" fn(i64) -> i64;
+/// The type of function pointer we'll return. We only support functions that take and return i64.
+type I64Func = extern "C" fn(i64) -> i64;
 
 /// A compiled module returned by `compile_module`, wrapping a `run` function that takes `i64`
 /// and returns `i64`. This structure includes (and manages) an LLVM execution engine, which is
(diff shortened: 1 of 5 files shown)


easy_ll/src/lib.rs
```diff
@@ -133,27 +139,27 @@ pub fn compile_module(code: &str, bc_file: Option<&[u8]>) -> Result<CompiledModu
...
+        check_run_function(module)?;
+        optimize_module(module)?;
 
         // Create an execution engine for the module and find its run function
-        let engine = try!(create_exec_engine(module));
+        let engine = create_exec_engine(module)?;
 
         result.engine = Some(engine);
-        result.function = Some(try!(find_run_function(engine)));
+        result.run_function = Some(find_function(engine, "run")?);
```
sppalkia: Can you add a note somewhere that the entry point is still a function called `run`?

weld_rt/src/lib.rs
```diff
@@ -269,3 +269,14 @@ pub extern "C" fn weld_rt_set_errno(run_id: libc::int64_t, errno: WeldRuntimeErr
     let entry = guarded.entry(run_id).or_insert(errno);
     *entry = errno;
 }
+
+#[no_mangle]
+/// Returns the current memory allocated to a run ID, or -1 if it is an invalid ID.
+pub extern "C" fn weld_rt_memory_usage(run_id: libc::int64_t) -> i64 {
```
sppalkia: Return `libc::int64_t` here

weld/lib.rs
```diff
@@ -180,12 +180,31 @@ pub unsafe extern "C" fn weld_value_free(obj: *mut WeldValue) {
...
+        let _ = module.run_named("rt_run_free", arg);
     }
     Box::from_raw(obj);
 }
 
 #[no_mangle]
+/// Gets the memory allocated to a Weld value, which generally includes all memory allocated during
+/// the execution of the module that created it (Weld does not currently free intermediate values
+/// while running a module). Returns -1 if the value is not part of a valid Weld run.
+pub unsafe extern "C" fn weld_value_memory_usage(obj: *mut WeldValue) -> i64 {
```
sppalkia: Return `libc::int64_t`

sppalkia:
Looks good, just a few small things.

easy_ll/src/lib.rs
mateiz: This is actually there in the comment above compile_module

PR #169, opened 2017-07-19
Eliminated shared runtime library

jjthomas:
Compile times may have gone up ... tests seem to take a bit longer to run. @mateiz maybe you can measure this?

`weld_module_free` and `weld_module_mem_free` are now used to free memory (they free all memory allocated by the module, which is what `weld_value_free` did before), and `weld_module_memory_usage` is used to determine a module's total memory usage. `weld_value_free` and `weld_value_memory_usage` now do nothing. I think we should implement these later once we figure out how to correctly free and determine the memory usage of a single value.


jjthomas:
Yep, that's right. I added some comments to the new functions clarifying this. The Python bindings have been updated.

README.md
```diff
@@ -41,6 +41,8 @@ $ ln -s /usr/local/bin/llvm-config-3.8 /usr/local/bin/llvm-config
 
 To make sure this worked correctly, run `llvm-config --version`. You should see `3.8.x` or newer.
 
+Enter the `weld_rt/cpp` directory and try running `make`. If the command fails with errors related to missing header files, you may need to install XCode and/or XCode Command Line Tools. Run `xcode-select --install` to do this.
```
mateiz: Is this a new step? I thought we had cargo run this and fail if it didn't work.

tests/integration_tests.rs
```diff
@@ -66,14 +67,16 @@ fn many_threads_conf() -> *mut WeldConf {
 unsafe fn _compile_and_run<T>(code: &str,
                               conf: *mut WeldConf,
                               ptr: &T)
-                              -> Result<*mut WeldValue, *mut WeldError> {
+                              -> Result<(*mut WeldValue, *mut WeldModule), *mut WeldError> {
```
mateiz: What does this return? Update the comment above

weld/lib.rs
```diff
@@ -312,14 +285,29 @@ pub unsafe extern "C" fn weld_module_run(module: *mut WeldModule,
...
+    let module = &mut *module;
+    module.run_named("rt_memory_usage", 0).unwrap()
+}
+
+#[no_mangle]
+/// Frees all unfreed memory allocated by this module over all prior runs.
+pub unsafe extern "C" fn weld_module_mem_free(module: *mut WeldModule) {
+    let module = &mut *module;
+    module.run_named("rt_run_free", 0).unwrap();
+}
```
mateiz: So is the idea that you call weld_module_mem_free to free stuff it allocated but leave the module around, and weld_module_free to delete the whole module? It seems a bit weird to tie memory management to a module. For example, can you call the same module from multiple threads now?

mateiz:
Basically people will want to compile a module once and then call it many times, possibly in different threads. Having only a module-level free function won't make sense for that.

README.md
jjthomas: It's not a new step, it's a one-time thing the user would do to check if they need to get XCode/XCode Command Line Tools.

jjthomas:
@mateiz @sppalkia fixed up the organization and naming of the runtime library, support for run_id is there as well

sppalkia:
@jjthomas for some reason on Travis, the `include_bytes!` call can't find `libweldrt` - is it being compiled before the rust stuff?

mateiz:
I'm in the process of debugging this here: https://github.com/weld-project/weld/pull/178. It's because the new runtime code depends on the C++ standard library, but we don't have that library loaded in Python on Linux for whatever reason. I think it's a dynamically linked library.

mateiz:
BTW the reason it did not build is because errno is a macro in C++ (!) so you can't have variables named errno: https://github.com/weld-project/weld/pull/178/commits/c262c9e34e4ed308d6fc8a7fcfa7912c0f924326

PR #175, opened 2017-07-20
Allow merging vectors into all builder types

### LINKED ISSUE #173: Support all Builders for Vectorization
Builders which do not have explicit vectorization should unroll the vector because merging values.

mateiz:
Also has some minor improvements for debug messages. Fixes #173.

### DIFF
--- a/weld/bin/repl.rs
+++ b/weld/bin/repl.rs
@@ -128,7 +128,7 @@ fn main() {
         }
 
         if let Err(ref e) = program {
-            println!("Error during parsing: {:?}", e);
+            println!("Error during parsing: {}", e);
             continue;
         }
 
--- a/weld/llvm.rs
+++ b/weld/llvm.rs
@@ -1593,7 +1593,7 @@ impl LlvmGenerator {
         // Special case: if the value is a SIMD vector, we need to merge each element separately (because the final
         // builder will still operate on scalars). We have specialized functions to merge all of them together for
         // some builders, such as Merger, but for others we just call the merge operation multiple times.
-        if let Simd(ref item) = *value_ty {
+        if vectorizer::is_simd(value_ty) {
             match *builder_kind {
                 Merger(ref t, ref op) => {
                     // For Merger, call the vectorMergePtr function to get a pointer to a vector we can merge into.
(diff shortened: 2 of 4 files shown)


weld/bin/repl.rs
```diff
@@ -128,7 +128,7 @@ fn main() {
         }
 
         if let Err(ref e) = program {
-            println!("Error during parsing: {:?}", e);
+            println!("Error during parsing: {}", e);
```
mateiz: This is just to prevent some weird escaped quotes from appearing here

weld/llvm.rs
```diff
@@ -1126,6 +1127,7 @@ impl LlvmGenerator {
 
     /// Generate code for a single statement, appending it to the code in a FunctionContext.
     fn gen_statement(&mut self, statement: &Statement, func: &SirFunction, ctx: &mut FunctionContext) -> WeldResult<()> {
+        ctx.code.add(format!("; {}", statement));
```
mateiz: This is amazingly useful -- look at the generated code after!

weld/vectorizer.rs
```diff
@@ -180,30 +199,27 @@ fn vectorizable(for_loop: &Expr<Type>) -> WeldResult<HashSet<Symbol>> {
...
-                                let mut getfield_passed = false;
-                                let ref elem_param = params[2];
-                                if let Ident(ref name) = expr.kind {
-                                    if elem_param.name == *name {
-                                       getfield_passed = true; 
-                                    }
-                                }
-                                passed = getfield_passed;
-                            }
-                            Merge{ .. } => {},
```
mateiz: @sppalkia is it bad to remove this? Getting rid of it allows more code to be vectorized, e.g. stuff that is creating intermediate structs for use in dictmerger, etc.

weld/llvm.rs
sppalkia: What does this do?

sppalkia: Ah never mind - it just appends the SIR expression

weld/vectorizer.rs
sppalkia: This should be okay...it works with nested structures and everything right? So basically, the behavior is to broadcast each literal in the struct?

mateiz:
Sure, in fact it did vectorize a lot of the existing tests because they failed without some fixes, but let me check for that explicitly.

weld/vectorizer.rs
```diff
@@ -14,12 +14,42 @@ use std::collections::HashSet;
...
+pub fn is_simd(ty: &Type) -> bool {
+    match *ty {
+        Simd(_) => true,
+        Struct(ref fields) => fields.iter().all(is_simd),
+        _ => false
+    }
+}
+
+/// Get the scalar type corresponding to a SIMD type.
+pub fn scalar_type(simd_type: &Type) -> WeldResult<Type> {
```
sppalkia: Btw, why is this here instead of being a method on `Type`? Same with `is_simd`

mateiz: Because `vectorized_type` was here. It may be OK to move them, but then we should move all 3 and we would probably want to make `vectorized_type` to return a WeldResult too to be consistent.

mateiz:
FYI this is failing because of the issue in #180 because we now vectorize the appender-based code in that test.

PR #176, opened 2017-07-20
Sequential loops

mateiz:
This implements sequential loops using the `iterate(initial_value, update_func)` construct, where `initial_value` is of some type `T` and `update_func` is of type `T => {T, bool}`. It works the same way as the sequential loop in NVL. We call `update_func` repeatedly on values starting from `initial_value` and stop when the `bool` it returns is `false`. Then the final value of the expression is the last `T` it produced.

I made this work slightly differently if the loop body is sequential vs parallel: in the sequential case, it just adds basic blocks in the current function, while in the parallel case, it adds new functions for the continuation, etc. This is the same way `If` generates code.

### DIFF
--- a/weld/sir.rs
+++ b/weld/sir.rs
@@ -937,7 +937,9 @@ fn gen_expr(expr: &TypedExpr,
             
             // After the body, unpack the {state, bool} struct into symbols argument_sym and continue_sym.
             let continue_sym = prog.add_local(&Scalar(ScalarKind::Bool), body_end_func);
-            prog.funcs[body_end_func].params.insert(argument_sym.clone(), initial.ty.clone());
+            if parallel_body {
+                prog.funcs[body_end_func].params.insert(argument_sym.clone(), initial.ty.clone());
+            }
             prog.funcs[body_end_func].blocks[body_end_block].add_statement(
                 GetField { output: argument_sym.clone(), value: result_sym.clone(), index: 0 });
             prog.funcs[body_end_func].blocks[body_end_block].add_statement(
--- a/weld/sir.rs
+++ b/weld/sir.rs
@@ -284,11 +284,11 @@ impl fmt::Display for Statement {
             Merge {
                 ref builder,
                 ref value,
-            } => write!(f, "merge {} {}", builder, value),
+            } => write!(f, "merge({}, {})", builder, value),
             Res {
                 ref output,
                 ref builder,
-            } => write!(f, "{} = result {}", output, builder),
+            } => write!(f, "{} = result({})", output, builder),
             NewBuilder {
                 ref output,
                 ref arg,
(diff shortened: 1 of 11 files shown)


mateiz:
@jjthomas you may want to think about whether the way this generates functions works with parlib. Basically you can now get some cycles in the function call graph. If you have code of the form `iterate(v, |v| { some for expression })`, then you will get a function graph like this:
```
F1:
  initial setup
  jump to F2 to start iterate() body

F2:
  invoke parallel for with body F3 and continuation F4

F3:
  body of parallel for

F4:
  if the bool in the result of iterate() is true
    jump to F1
  else
    return final result
```

jjthomas:
@mateiz yep I think this will work fine. Once parallel loop inside the sequential loop finishes executing (continuation completes), there is no state left and it can be run again. I will think more but there shouldn't be problems.

mateiz:
Yeah, the only tricky thing I found in writing it was that I had to add some variables explicitly as parameters of the nested functions to thread them through to F1/F2/F3 in that example.

weld/sir.rs
```diff
@@ -876,6 +882,89 @@ fn gen_expr(expr: &TypedExpr,
...
+
+            // Generate the loop's body, which will work on argument_sym and produce result_sym.
+            // The type of result_sym will be {ArgType, bool} and we will repeat the body if the bool is true.
+            let (body_end_func, body_end_block, result_sym) =
+                gen_expr(func_body, prog, body_start_func, body_start_block)?;
+            
+            // After the body, unpack the {state, bool} struct into symbols argument_sym and continue_sym.
+            let continue_sym = prog.add_local(&Scalar(ScalarKind::Bool), body_end_func);
+            if parallel_body {
+                prog.funcs[body_end_func].params.insert(argument_sym.clone(), initial.ty.clone());
```
jjthomas: Why do you need this line?

weld/sir.rs
```diff
@@ -876,6 +882,89 @@ fn gen_expr(expr: &TypedExpr,
...
+                Assign { output: argument_sym.clone(), value: initial_sym });
+
+            // Check whether the function's body contains any parallel loops. If so, we should put the loop body
+            // in a new function because we'll need to jump back to it from continuations. If not, we can just
+            // make the loop body be another basic block in the current function.
+            let parallel_body = contains_parallel_expressions(func_body);
+            let body_start_func = if parallel_body {
+                let new_func = prog.add_func();
+                //prog.add_local_named(&initial.ty, argument_sym, new_func);
+                prog.funcs[new_func].params.insert(argument_sym.clone(), initial.ty.clone());
```
jjthomas: Or this line?

jjthomas:
Ah just saw your comment where you mention this, could you explain more why the automatic argument correction didn't work? `argument_sym` seems to be used both as a param and as an output value in some functions.

mateiz:
Yup, it is used that way because we overwrite it as we update the result for the iteration. The normal parameter helper thing didn't add it as a parameter, so I had to add it myself. Not sure why it didn't add it.

jjthomas:
Looked at this a little more closely and I think `argument_sym` won't always get threaded through the chain of functions because `sir_param_correction` only considers variables consumed in expressions, and `argument_sym` is not consumed in the final function but is instead assigned. (In the integration test since `argument_sym` is consumed in the `map` and there are no further loops in the continuation, everything works out fine.) I will try to fix this tomorrow.

mateiz:
OK, should I still merge it or wait for you to fix?

PR #177, opened 2017-07-20
Refactor of `llvm.rs` Part 1

sppalkia:
This pull requests cleans up/refactors `llvm.rs` to be more readable. The status of the refactor will be updated here.

### Finished

* Break up `llvm_type` into several functions
* Consistent function names (e.g., all functions that generate code have the prefix `gen_`)
* Remove redundant patterns (e.g., instead of always calling `llvm_type(..)?.to_string()`, `llvm_type` now returns `WeldResult<String>` instead of `WeldResult<&str>`
* Add function to generate stores (`gen_store_var`)
* Use `?` instead of `try!` everywhere
* `replace!` macro to replace tokens in templates (makes code more concise
* Added a function `llvm_type_and_name` to get commonly used symbol information (LLVM type string, LLVM symbol formatted as string)
* `get_sym_ty` is renamed to `symbol_type` and is a method on `SirFunction` now

### To do in a Future PR

* The `add_function` method needs a major overhaul - I haven't started this yet.
* Some naming convention (maybe even an enum with strings so Rust type checks for us) for registers, pointers, etc.


weld/llvm.rs
```diff
@@ -38,6 +38,22 @@ static DICTIONARY_CODE: &'static str = include_str!("resources/dictionary.ll");
 static DICTMERGER_CODE: &'static str = include_str!("resources/dictmerger.ll");
 static GROUPMERGER_CODE: &'static str = include_str!("resources/groupbuilder.ll");
 
+/// Replace occurances of each `key` in `template` with `value`.
+///
+/// // returns "My name is Shoumik"
+/// replace!("My name is $NAME", "$NAME" => "shoumik")
```
mateiz: Wouldn't it be better to use `format!` for this? Maybe we'd have to change our templates a bit but it cloud be more consistent, especially since some .ll templates do use the `format!` format.

weld/llvm.rs
```diff
@@ -765,223 +777,279 @@ impl LlvmGenerator {
...
-
-            Simd(Bool) => Ok(self.simd_names.entry(Bool).or_insert(format!("<{} x i1>", vec_size(&Scalar(Bool))?))),
-            Simd(I8) => Ok(self.simd_names.entry(I8).or_insert(format!("<{} x i8>", vec_size(&Scalar(I8))?))),
-            Simd(I32) => Ok(self.simd_names.entry(I32).or_insert(format!("<{} x i32>", vec_size(&Scalar(I32))?))),
-            Simd(I64) => Ok(self.simd_names.entry(I64).or_insert(format!("<{} x i64>", vec_size(&Scalar(I64))?))),
-            Simd(F32) => Ok(self.simd_names.entry(F32).or_insert(format!("<{} x float>", vec_size(&Scalar(F32))?))),
-            Simd(F64) => Ok(self.simd_names.entry(F64).or_insert(format!("<{} x double>", vec_size(&Scalar(F64))?))),
+    fn llvm_type(&mut self, ty: &Type) -> WeldResult<String> {
+        Ok(match *ty {
+            Scalar(kind) => llvm_scalarkind(kind).to_string(),
```
mateiz: Use snake case (`llvm_scalar_kind`)

mateiz:
BTW it would also be nice to have a standard naming convention for LLVM type vs Weld type, register vs stack variable, etc. Not sure whether you can easily do that here.

weld/llvm.rs
```diff
@@ -765,223 +781,279 @@ impl LlvmGenerator {
...
+    //
+    *********************************************************************************************/
+
+    /// Returns the Type, LLVM type (as a string), and LLVM symbol name (as a string).
+    fn symbol_info<'a>(&mut self, func: &'a SirFunction, sym: &Symbol) -> WeldResult<(&'a Type, String, String)> {
+        let ty = func.symbol_type(sym)?;
+        let llvm_ty = self.llvm_type(ty)?;
+        let llvm_str = llvm_symbol(sym);
+        Ok((ty, llvm_ty, llvm_str))
+    }
```
mateiz: FYI I'm not sure that this is clearer than what we had before. I see a lot of `let (_, blah_ty, blah_sym) = symbol_info(...)`, which is pretty hard to parse -- someone unfamiliar with the code would not know what those are until they read the doc for `symbol_info`. It also seems weird to return 3 values and then throw one away. I think you can fix this in one of the following ways: either (1) return an Info object that has type, symbol, etc as fields, or (2) just use separate functions for each thing -- after all, I see no efficiency arguments in computing all three of these at once.

mateiz: Or if we *always* get the last two fields and only sometimes get the first one, maybe just make this return those two fields, and call `symbol_type` when you need the other one. Then you should also pick a more descriptive name than `symbol_info`, maybe `llvm_name_and_type` or `llvm_info` or something.

sppalkia: Okay, I changed this to `llvm_type_and_name`, and then you call `symbol_type` explicitly if you need the type. It's a bit weird since `llvm_type_and_name` needs to call `symbol_type` too, but it's more readable.

sppalkia: btw the reason I made this change was that it took 3-4 lines of boilerplate code to get the name/type for *each* symbol, so it was making the code really long/messy.

sppalkia:
@mateiz should be ready for merge - still some stuff to do but `gen_statement` should be cleaned up.

mateiz:
@sppalkia  it looks like you never addressed one of my comments, about the `replace!` macro: "Wouldn't it be better to use `format!` for this? Maybe we'd have to change our templates a bit but it cloud be more consistent, especially since some .ll templates do use the `format!` format."

sppalkia:
Oops sorry - this has been addressed.

PR #178, opened 2017-07-21
Move runtime library to C++ and allow parallel calls to same module

mateiz:
This adds a few changes to #169 to build on it


weld/util.rs
```diff
@@ -109,16 +107,3 @@ pub fn get_weld_home() -> Result<String, ()> {
         Err(_) => Err(()),
     }
 }
-
-/// Loads the Weld runtime library.
-pub fn load_runtime_library() -> Result<(), String> {
-    let weld_home = get_weld_home().unwrap_or("./".to_string());
-    let path = format!("{}{}", weld_home, "weld_rt/target/release/libweldrt");
```
sppalkia: Hooray! This was one of the messiest parts of the run process.

weld_common/src/lib.rs
```diff
@@ -5,14 +5,14 @@ use std::fmt;
...
-    Success = 0,
+    Success = 0, // explicit values for Success and OutOfMemory required due to use by weld_rt C++ code
     ConfigurationError,
     RuntimeLibraryError,
     CompileError,
     ArrayOutOfBounds,
     BadIteratorLength,
     MismatchedZipSize,
-    OutOfMemory,
+    OutOfMemory = 7,
```
sppalkia: Why is the explicit value required here? Aren't these guaranteed to be sequential?

mateiz: In case people insert more of them above, sort the file, etc. It makes sense to keep it.

PR #179, opened 2017-07-21
Fixes issue with Vectorization and work splitting

sppalkia:
This fixes an issue with vectorization and work splitting. Before, work splitting could cause task sizes that were not multiples of the SIMD vector size; this would cause vectorization to miss elements/compute incorrect results, since it assumes all fringing occurs at the _end_ of the loop.

This fixes the issue by making sure the split-point for tasks is always a divisible of the SIMD size (currently, the maximum assumed SIMD size is 64 bytes). An additional check has been added to `parlib.cpp` when splitting: if the `grain_size` is greater than some threshold, the above adjustment is made to the split point for tasks. Otherwise, the adjustment is not made because the task is assumed to be a split for an outer loop.

@jjthomas can you make sure these changes are safe?

### DIFF
--- a/examples/cpp/vector_benchmark/vector_benchmark.cpp
+++ b/examples/cpp/vector_benchmark/vector_benchmark.cpp
@@ -17,7 +17,8 @@ struct args {
     int32_t a;
 };
 
-const char *program = "|x:vec[i32], a:i32| result(for(x, merger[i32,+], |b,i,e| merge(b, e+a)))";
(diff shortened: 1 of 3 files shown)


sppalkia:
Will re-open this on top of #178.

mateiz:
FYI you can just `git rebase` your branch onto it and then `git push -f` to update the same PR.

PR #181, opened 2017-07-21
fixed code generation for Iterate corner cases

jjthomas:
@mateiz can you take a look, your `iterate_with_parallel_body` is basically a test for this (it won't pass anymore without the change to `sir_param_correction_helper`

### DIFF
--- a/tests/integration_tests.rs
+++ b/tests/integration_tests.rs
@@ -1585,7 +1585,7 @@ fn iterate_non_parallel() {
 }
 
 fn iterate_with_parallel_body() {
-    let code = "|x:i32| iterate({[1,2,3], 1}, |p| {{map(p.$0, |y|y*2), p.$1+1}, p.$1<x}).$0";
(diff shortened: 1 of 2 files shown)


jjthomas:
never mind, updated `iterate_with_parallel_body` so that it fails without this change

mateiz:
Seems OK, but can you make sure it's not adding more parameters to the closure than needed in other cases? I guess that in normal cases, those have unique IDs? It would be nice to directly test sir_param_correction actually -- just make a few cases and see what it does.

jjthomas:
It definitely can't add more params to the closure that needed because anything that's already in the params list of a called function must be passed in by its caller, so we are safe there. I can add tests for `sir_param_correction` if you want.

mateiz:
OK, this fix seems fine but it would be good to add a full test to this that just creates a SIR program and then runs this (or even that goes from AST to SIR).

PR #183, opened 2017-07-23
Allow dictmerger values to be structs

mateiz:
This changes `dictmerger.ll` slightly to call a function for each merge instead of assuming it's one instruction. I also added a test for this.


deepakn94:
Looks good to me.

mateiz:
OK, feel free to merge it if it passes (it's the end of the day here).

PR #184, opened 2017-07-24
uniquifier bugfix & better iterate codegen to reduce extraneous params

jjthomas:
@sppalkia check the uniquifier fix and see if it looks good

### DIFF
--- a/tests/integration_tests.rs
+++ b/tests/integration_tests.rs
@@ -360,6 +360,21 @@ fn comparison() {
     unsafe { free_value_and_module(ret_value) };
 }
 
+fn reused_variable() {
+    // `a` is reused in different scopes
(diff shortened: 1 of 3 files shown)


jjthomas:
actually hold off on merging this, i am going to make the tests a little better

jjthomas:
@sppalkia @mateiz take a look at this now, the new tests actually didn't really test anything interesting so i removed them

mateiz:
Wait so what is this fixing then? Can't we add a test that reproduces the problem?

jjthomas:
@mateiz pushed a test that breaks without the uniquifier fix, the other change to sir.rs just improves the generated code so it doesn't pass around unneeded params, so there's not a test for it

mateiz:
Awesome, thanks!

PR #185, opened 2017-07-24
[WIP] Faster hashing and equality functions

mateiz:
This makes a few changes that improve performance in workloads with hashing of multi-field structures:

* Change `hash_combine` method to `(31 * seed) + value`, which is faster than what we had so far and is used by lots of Java classes, etc. I also added a `hash_finalize` step to the dictionary code that mixes the bits within this more.
* Change hash functions to 32 bits instead of 64 -- this made a ~10% difference and there is no reason to have 64 bits here.
* Add an explicit equality function to all types that can be faster than their `cmp` function (by not branching as often as `cmp`).

### DIFF
--- a/weld/llvm.rs
+++ b/weld/llvm.rs
@@ -813,6 +813,7 @@ impl LlvmGenerator {
                     self.prelude_code.add_line(format!("}}"));
                     self.prelude_code.add_line(format!(""));
 
+                    // Generate a comparison function for the struct
                     self.prelude_code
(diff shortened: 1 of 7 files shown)


jjthomas:
Looks fine, seems like the new equality function could be slower in some cases, I guess we are expecting that we will mostly compare things are equal in the dictionary case?

mateiz:
@jjthomas OK, I actually added back the branches in `eq` now because removing those didn't seem to have much impact on performance. Switching from `cmp` to `eq` does though.

PR #186, opened 2017-07-25
Numpy

Parimarjan:
First implementation for 1d numpy operations - most of the things to review are in weldarray.py.


python/numpy/timings/black_scholes_weld.py
```diff
@@ -0,0 +1,111 @@
...
+import random
+
+invsqrt2 = 1.0 #0.707
+verbose = False
+
+os.environ["NUM_WORKERS"] = "12"
+
+import black_scholes_config as bsc
+
+def black_scholes_nvl(price, strike, t, rate, vol):
```
mateiz: Why are you calling it NVL?

Parimarjan: that had just been left over from the old version, I'll make the change to the names in this file

python/numpy/main.py
```diff
@@ -0,0 +1,28 @@
+from weldarray import WeldArray
```
sppalkia: I'd get rid of this file before merging this PR in, just to keep the repo clean.

python/numpy/timings/black_scholes_config.py
```diff
@@ -0,0 +1,21 @@
+#
+# Configuration for the Black Scholes workload.
```
sppalkia: Same here - These things should probably go in their own repo under weld-project instead of being here.

python/numpy/weldarray.py
```diff
@@ -0,0 +1,375 @@
...
+            return result
+
+    def evaluate(self):
+        '''
+        User facing function - if he wants to explicitly evaluate all the
+        registered ops.
+        '''
+        return WeldArray(self._eval())
+
+    def _eval(self):
```
sppalkia: We should also implement caching  here -- e.g., if we print out the same array twice, we shouldn't evaluate it twice if the value doesn't change

python/numpy/main.py
Parimarjan: makes sense, removed this file

python/numpy/timings/black_scholes_config.py
Parimarjan: removed all the black scholes files

python/numpy/weldarray.py
Parimarjan: actually it does do caching I think - every time _eval is called, and there are registered ops, it evaluates them, and creates a new weldobject with no ops registered, so next time self.name == weldobj.weld_code is true because of the way I'd set it up. maybe I should make this more explicit?

sppalkia: Ah, didn't catch that. I think it's okay the way you have it :)

python/numpy/weldarray.py
```diff
@@ -0,0 +1,375 @@
...
+        @result: None or a weldarray to store results in.
+        '''
+        if result is None:
+            result = WeldArray(self)
+        else:
+            # FIXME: deal with updating view children/parents
+            pass
+
+        # FIXME: support other dtypes - not sure what to do for f64/or i64.
+        if isinstance(other, np.float32) or isinstance(other, float):
```
sppalkia: Better way to do this might be to have some dictionary called `DTYPE_SUFFIXES`, which maps every value in `SUPPORTED_DTYPES` to a string (e.g., `""` for `i32`, `"f"` for `f32`, etc.). That way, when we add new types we don't have to go through all the code and find the points where support for the type needs to be added (then here, you would just index into the dictionary with the `dtype`.

You could even make a class called `DTypeInfo` or something and make this suffix a field in that object, so we need to add other type specific stuff we can append things to that object.

python/numpy/weldarray.py
```diff
@@ -0,0 +1,375 @@
...
+        result.weldobj.weld_code = code
+
+        if not is_weld_array:
+            # Evaluate all the lazily stored ops, as the caller probably
+            # expects to get back a ndarray - which would have correct values.
+            result = WeldArray(result._eval())
+
+        return result
+
+# TODO: This should finally go into weldnumpy - along with other functions
```
sppalkia: See note on reorganizing this a bit

sppalkia:
@mateiz and @deepakn94 I think this is good to go, and I'd like to use it for the DAWN demo - can you guys leave any other comments you might have for @Parimarjan? Otherwise, we can merge it in.

mateiz:
If you think it's good, merge it in -- it will be easier to work based on it there. Just make sure it doesn't complicate the build system and publishing stuff we're doing.

deepakn94:
Looks pretty good -- I propose a slight directory reorg (to match the `grizzly` organization) -- why not move everything in `numpy/weldnumpy/` to `numpy`. You can keep the `tests` and `timing` directories

Parimarjan:
@deepakn94, @sppalkia had suggested using this directory structure - do you guys think it should be changed back to the previous structure without weldnumpy?

sppalkia:
Let's keep it as is for now and we can change it later -- like Matei said, it'll be easier for everyone to make changes to it when its in master.

PR #189, opened 2017-07-25
Implement faster comparison for vec[u8] using memcmp

mateiz:


weld/llvm.rs
```diff
@@ -901,7 +903,22 @@ impl LlvmGenerator {
                     self.prelude_code.add(&name_replaced);
                     self.prelude_code.add("\n");
 
-                    // Supports vectorization, so splice in the vector extensions.
+                    // Generate comparison functions, using a special template for vec[i8].
+                    if *elem.as_ref() == Scalar(ScalarKind::I8) {
```
sppalkia: Is there a reason why this only works with `i8` instead of any scalar integer type?

mateiz: It won't be correct for signed integers. Probably won't work well on floats/doubles either but I'm not 100% sure.

sppalkia: Ah, because it's overriding the lexicographic comparison function. It might make more sense to add an explicit equality function for `==`, so we can just do a byte-level comparison for any scalar-based vector.

mateiz: Yeah, that's true, that's one reason I separated "eq" from "cmp". Even then though, it's technically incorrect for floats because NaN != NaN; we need to see how we want to handle that (but I believe our built in comparisons do the right thing there).

mateiz:
Just updated this to work with the new unsigned types.

sppalkia:
(repoend to run travis again)

mateiz:
Closing to rerun travis

PR #190, opened 2017-07-25
Add a weld_load_library function for loading shared libraries

mateiz:
This is useful when loading C UDFs in Java / Python, where they are not already linked into the program.


jjthomas:
Looks good

PR #194, opened 2017-08-03
Removing stdlib=libstdc++ from Makefile

Parimarjan:
Fixes llvm linking issue (LLVM ERROR: Program used external function
'__ZSt28_Rb_tree...') for my osx, El Capitan, 10.11.6, (libc++ seems
to be the default on it)


jjthomas:
@mateiz can you verify this works for you

mateiz:
Yup, seems to work. I'll merge it.

PR #199, opened 2017-08-11
Switch logging to use the log crate and add it to more places

mateiz:
This allows us to get fine-grained logging of each step in LLVM, code generation, etc without having to pass our own Log object around. The `log` crate was proposed as a standard facade to logging libraries in Rust (https://doc.rust-lang.org/log/log/index.html).

### DIFF
--- a/c/weld.h
+++ b/c/weld.h
@@ -9,6 +9,18 @@
 
 #include <stdint.h>
 
+// Log level constants -- should match WeldLogLevel.
+
+/** A Weld log level. */
+typedef uint64_t weld_log_level_t;
+
+weld_log_level_t WELD_LOG_LEVEL_OFF = 0;
+weld_log_level_t WELD_LOG_LEVEL_ERROR = 1;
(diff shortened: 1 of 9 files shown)


c/weld.h
```diff
@@ -9,6 +9,18 @@
 
 #include <stdint.h>
 
+// Log level constants -- should match WeldLogLevel.
+
+/** A Weld log level. */
+typedef uint64_t weld_log_level_t;
+
+weld_log_level_t WELD_LOG_LEVEL_OFF = 0;
```
sppalkia: Why not make this an enum?

mateiz: I wasn't sure the enum would be guaranteed to be 64-bit. How can you force it to be that?

sppalkia: Hmm yeah, I don't think that can be guaranteed. You could make this an enum and then cast the value you get passed from rust to the enum type, but maybe this is better then.

sppalkia:
Looks good otherwise!

PR #210, opened 2017-08-19
removed thread creation for num_workers=1

jjthomas:
@mateiz will run some performance benchmarks tomorrow

### DIFF
--- a/weld_rt/cpp/Makefile
+++ b/weld_rt/cpp/Makefile
@@ -1,7 +1,7 @@
 OS = $(shell uname -s)
 LLVM_VERSION = $(shell llvm-config --version | cut -d . -f 1,2)
 
-CFLAGS = -O3 -std=c++11 -Wall -fno-use-cxa-atexit -fPIC -flto
+CFLAGS = -O0 -g -std=c++11 -Wall -fno-use-cxa-atexit -fPIC -flto
 ifeq (${OS}, Darwin)
   # OS X
   CLANG ?= clang++-${LLVM_VERSION}
--- a/weld_rt/cpp/runtime.cpp
+++ b/weld_rt/cpp/runtime.cpp
@@ -7,6 +7,7 @@
 #include <queue>
 #include <deque>
 #include <algorithm>
+#include <exception>
 #include "assert.h"
 #include "runtime.h"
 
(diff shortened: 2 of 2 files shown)


jjthomas:
Ah yeah as the Travis failure shows on systems there is some code that checks that the `pthread_exit` exception was thrown all the way up the stack, so I'm not sure if this solution will work ...

deepakn94:
@jjthomas, maybe you can use weld-benchmarks when you're running the performance benchmarks?

jjthomas:
@deepakn94 this was just to measure the cost of creating a pthread, but I will definitely use them next time

weld_rt/cpp/Makefile
```diff
@@ -1,7 +1,7 @@
 OS = $(shell uname -s)
 LLVM_VERSION = $(shell llvm-config --version | cut -d . -f 1,2)
 
-CFLAGS = -O3 -std=c++11 -Wall -fno-use-cxa-atexit -fPIC -flto
+CFLAGS = -O0 -g -std=c++11 -Wall -fno-use-cxa-atexit -fPIC -flto
```
sppalkia: Need to change this back (just a reminder so we don't forget)

weld_rt/cpp/runtime.cpp
```diff
@@ -335,22 +338,26 @@ static void *thread_func(void *data) {
...
-      iters = 0;
-      work_loop(td->thread_id, rd);
-    } else {
-      // If this thread is stalling, periodically check for errors.
-      iters++;
-      if (iters > 1000000) {
-        if (rd->err != 0) {
-	  break;
-        }
+  try {
```
sppalkia: Does this add a bunch of overhead? We should benchmark it

sppalkia: According to this it shouldn't impose too much overhead, just increase binary size: https://stackoverflow.com/questions/16784601/does-try-catch-block-decrease-performance so we're probably okay

jjthomas:
Yeah exceptions have zero performance overhead just code size

weld_rt/cpp/Makefile
jjthomas: Yeah this is required for now, hopefully when @mateiz converts this to a shared library the problem goes away and we can revert to -O3. It shouldn't be a big problem even now because MCJIT further optimizes this code

sppalkia:
Got it - you can merge this if it works, looks good.

PR #213, opened 2017-08-31
fast path for fixed-size Appender

jjthomas:
@sppalkia can you test performance


sppalkia:
```
C Runtime: 0.031014
Weld Runtime: 0.034031
```

beautiful job good sir

PR #214, opened 2017-09-01
Predicate only when annotation is true

pratiksha:
Plus, refactoring to add annotations to If and For exprs.

This change means predication will never be applied to code that doesn't have the annotation -- I don't think this breaks backwards-compatibility since predication is turned off by default anyway, but let me know.


deepakn94:
Looks good to me.
Can you also add an end-to-end test similar to [this](https://github.com/weld-project/weld/blob/master/tests/integration_tests.rs#L880), but with an `if` condition that both has and doesn't have the `predicate` annotation?

pratiksha:
@deepakn94 ptal

deepakn94:
LGTM, ship it! (Use the `squash and merge` option)

deepakn94:
(Wait for tests on Travis to pass first)

PR #215, opened 2017-09-02
Add unsigned and 16-bit integer types

mateiz:


weld/ast.rs
```diff
@@ -111,20 +111,79 @@ impl Type {
...
+            I8 | I16 | I32 | I64 => true,
+            _ => false
+        }
+    }
+
+    /// Is this type a signed integer of any type (excludes Bool)?
+    pub fn is_unsigned_integer(&self) -> bool {
+        use ast::ScalarKind::*;
+        match *self {
+            U8 => true,
```
sppalkia: Should we also add `u16`, `u32`, and `u64` while we're at it?

mateiz: Good point, might as well add them

PR #216, opened 2017-09-03
Refactor weld_common/ and easy_ll/ crates into modules

deepakn94:
This is to ensure that we can publish Weld as a single crate.


tests/integration_tests.rs
```diff
@@ -2,10 +2,9 @@ use std::env;
 use std::thread;
 
 extern crate weld;
-extern crate weld_common;
 extern crate libc;
 
-use weld_common::WeldRuntimeErrno;
+use weld::weld_common::WeldRuntimeErrno;
```
mateiz: should we rename this module to `common` or just put it in the main weld `lib.rs`?

weld/easy_ll/tests.rs
```diff
@@ -15,7 +16,7 @@ fn basic_use() {
            %2 = call i64 @bar(i64 %1)
            ret i64 %2
        }
-    ");
+    ", Some(LIB_WELD_RT));
```
mateiz: You don't need to pass this in the easy_ll tests, do you? isn't it enough to pass None?

mateiz:
Looks OK overall but I added a couple of comments

weld/easy_ll/tests.rs
deepakn94: Fixed.

deepakn94:
@mateiz Addressed your comments.

tests/integration_tests.rs
deepakn94: Renamed the module to `common`

PR #217, opened 2017-09-03
switched from exceptions to setjmp/longjmp for error exiting

jjthomas:


weld_rt/cpp/runtime.cpp
```diff
@@ -322,15 +338,17 @@ static inline void work_loop(int32_t my_id, run_data *rd) {
...
+	cleanup_tasks_on_thread(popped, my_id, rd);
         return;
       }
       if (!setjmp(rd->work_loop_roots[my_id])) {
         popped->fp(popped);
+        finish_task(popped, my_id, rd);
       } else {
-	// TODO we should clean up this task as well as other tasks that might be left on work queues
+	// error-case exit from task
+	cleanup_tasks_on_thread(popped, my_id, rd);
```
mateiz: Formatting seems to be off?

jjthomas: think this is fixed, thanks

mateiz:
Looks good, I've merged it.

PR #218, opened 2017-09-04
Build native library just once and various updates to reduce compile times

mateiz:
* Builds most of the C++ runtime library into a `libweldrt` instead of rebuilding it each time you compile a Weld module (which took about 500 ms with O2 optimizations). A few key functions are still saved as a .bc file to be inlined.
* Makes the LLVM optimization level configurable.
* Changes the way we generate code for structs to reduce the number of temporary structs stored in registers, which can slow down / confuse LLVM codegen. We use `getelementptr` on the struct to load the required field directly instead. (This can give a 20% compile time speedup).

### DIFF
--- a/weld/conf.rs
+++ b/weld/conf.rs
@@ -98,7 +98,7 @@ fn parse_passes(s: &str) -> WeldResult<Vec<Pass>> {
 /// Parse an LLVM optimization level.
 fn parse_llvm_optimization_level(s: &str) -> WeldResult<u32> {
     match s.parse::<u32>() {
-        Ok(v) if v <= 2 => Ok(v),
+        Ok(v) if v <= 3 => Ok(v),
         _ => weld_err!("Invalid LLVM optimization level: {}", s),
     }
 }
--- a/weld/conf.rs
+++ b/weld/conf.rs
@@ -11,10 +11,12 @@ use super::passes::Pass;
 pub const MEMORY_LIMIT_KEY: &'static str = "weld.memory.limit";
 pub const THREADS_KEY: &'static str = "weld.threads";
 pub const OPTIMIZATION_PASSES_KEY: &'static str = "weld.optimization.passes";
+pub const LLVM_OPTIMIZATION_LEVEL_KEY: &'static str = "weld.llvm.optimization.level";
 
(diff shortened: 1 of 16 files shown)


weld/conf.rs
```diff
@@ -86,17 +95,31 @@ fn parse_passes(s: &str) -> WeldResult<Vec<Pass>> {
     Ok(result)
 }
 
+/// Parse an LLVM optimization level.
+fn parse_llvm_optimization_level(s: &str) -> WeldResult<u32> {
+    match s.parse::<u32>() {
+        Ok(v) if v <= 2 => Ok(v),
```
sppalkia: Maybe make this 4 or something? Since LLVM also supports O2 and O3.

We may (eventually) want to make our own passes too, that e.g., disable loop vectorization and stuff since that will never happen.

weld/lib.rs
```diff
@@ -242,6 +242,15 @@ pub unsafe extern "C" fn weld_module_compile(code: *const c_char,
     let code = CStr::from_ptr(code);
     let code = code.to_str().unwrap().trim();
 
+    // Let LLVM find symbols in libweldrt; it's okay to call this multiple times
+    let libweldrt = if cfg!(target_os="macos") {
+        "libweldrt.dylib"
+    } else {
+        "libweldrt.so"
```
sppalkia: Add a DLL for windows? Not sure if the rest of the code even works with Windows atm though

weld/conf.rs
mateiz: It actually only takes up to 3, see http://llvm.org/doxygen/classllvm_1_1PassManagerBuilder.html#a52b96ef67bb93fa3a7b461fe2b73bb4f. We should add other configs if we want other sets of passes.

jjthomas:
Looks good, you didn't run into any of the previous issues we had when linking weldrt into the main libweld (unused symbols getting removed, etc.)?

mateiz:
No, it stays linked dynamically, and the C++ compiler doesn't drop any of those symbols.

sppalkia:
Just checked the assembly/optimized LLVM code and it looks good - you can merge it!

PR #219, opened 2017-09-04
bug fix for partial test matching

pratiksha:
Match all tests containing input string.


sppalkia:
Looks good, thanks! I broke this because I thought it was a bug not a
feature haha

PR #222, opened 2017-09-04
logging updates - adding llvm IR code to trace log

Parimarjan:


weld/easy_ll/mod.rs
```diff
@@ -109,6 +109,26 @@ pub fn load_library(libname: &str) -> Result<(), LlvmError> {
     }
 }
 
+fn output_llvm_ir(engine: LLVMExecutionEngineRef, module: LLVMModuleRef) -> String {
+    // Keep enough space for the llvm IR being created.
+    let mut v = [8 as i8; 200000];
```
sppalkia: Is there any way to check the size to allocate before instead of hard coding it? This may not be big enough for giant programs.

weld/easy_ll/mod.rs
```diff
@@ -155,6 +175,7 @@ pub fn compile_module(code: &str, bc_file: Option<&[u8]>) -> Result<CompiledModu
         // Create an execution engine for the module and find its run function
         let engine = create_exec_engine(module)?;
         debug!("Done creating execution engine");
+        trace!("llvm ir for target:\n = {}", output_llvm_ir(engine, module));
```
sppalkia: Maybe call it "Optimized LLVM IR:"

sppalkia:
(reopened to run travis again)

weld/easy_ll/mod.rs
Parimarjan: this was some stuff I left here by mistake...we just need an arbitrary pointer here which would be changed by llvm...have fixed this in the latest commit.

python/weld/bindings.py
```diff
@@ -46,6 +46,7 @@ class c_weld_value(c_void_p):
 class WeldModule(c_void_p):
 
     def __init__(self, code, conf, err):
+        # weld.weld_set_log_level(5)
```
sppalkia: Delete

python/weld/weldobject.py
```diff
@@ -130,7 +130,7 @@ def toWeldFunc(self):
             "\n".join(list(set(self.constants))) + "\n" + self.weld_code
         return text
 
-    def evaluate(self, restype, verbose=True, decode=True):
+    def evaluate(self, restype, verbose=True, decode=True, log_level=0):
```
sppalkia: Maybe we should get rid of verbose? Or change it to the following:

verbose=False is log level 0
verbose=True is log level info
otherwise if verbose=n, log level = n

weld/easy_ll/mod.rs
```diff
@@ -109,6 +111,35 @@ pub fn load_library(libname: &str) -> Result<(), LlvmError> {
...
+    // modified by EmitToMemoryBuffer.
+    let mut output_buf : self::llvm::prelude::LLVMMemoryBufferRef = ptr::null_mut();
+    let mut err = ptr::null_mut();
+    unsafe {
+        let cur_target = LLVMGetExecutionEngineTargetMachine(engine);
+        let file_type = LLVMCodeGenFileType::LLVMAssemblyFile;
+        let res = LLVMTargetMachineEmitToMemoryBuffer(cur_target, module, file_type, &mut err, &mut output_buf);
+        if res == 1 {
+            let x = CStr::from_ptr(err as *mut c_char);
+            panic!("Getting LLVM IR failed! {:?}", x);
```
sppalkia: Don't panic here: change the return type to WeldResult<String> and return an Err()

sppalkia:
@Parimarjan can you fix the conflicts? Ill merge this after. Thanks!

sppalkia:
Hey Pari,

I added some other stuff that does this, so I'm going to close this PR for now.

PR #223, opened 2017-09-04
Fix boolean equality

hvanhovell:
`Equals` && `NotEquals` expressions are currently broken for boolean inputs. For example the following program:
```
|b: bool| b == false
```
...fails with the following error:
```
Unsupported binary op: == on bool
```
This issue was introduced by https://github.com/weld-project/weld/commit/7bf6469e8f018ffdea418c63d9e1051877b641b8. This PR fixes this issue by adding the required checks to `llvm.rs`.

### DIFF
--- a/tests/integration_tests.rs
+++ b/tests/integration_tests.rs
@@ -208,6 +208,27 @@ fn negated_arithmetic() {
     unsafe { free_value_and_module(ret_value) };
 }
 
+fn bool_eq() {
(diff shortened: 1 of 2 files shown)


hvanhovell:
cc @mateiz @sppalkia

hvanhovell:
Hmmm something is up with travis...

sppalkia:
(reopened to run travis again)

PR #224, opened 2017-09-05
Travis Fix

sppalkia:


sppalkia:
This temporarily removes LLVM 4.0, since it seems to be removed from `apt-get`.

PR #225, opened 2017-09-06
build wheel

rahulpalamuttam:
steps to build and publish test wheel:

Write ~/.pypirc file:
```
--------------------
[distutils]
index-servers =
   testpypi

[testpypi]
repository=https://test.pypi.org/legacy/
username=**
password=**
----------------------

```
Commands:

python setup.py sdist bdist_wheel
twine upload -r testpypi dist/*.whl

Installing the test wheel:
pip install --index-url https://test.pypi.org/simple/ weld


python/weld/bindings.py
```diff
@@ -8,17 +8,39 @@
...
     lib_file = "libweld.dylib"
+    librt_file = "libweldrt.dylib"
 else:
     raise OSError("Unsupported platform {}", system)
 
+lib_file = pkg_resources.resource_filename(__name__, lib_file)
+librt_file = pkg_resources.resource_filename(__name__, librt_file)
+
+lib_path, lib_filename = os.path.split(lib_file)
+os.chdir(lib_path)
```
rahulpalamuttam: I'm not sure if this is best practice - but it worked.

sppalkia:
Hey Rahul, can you fix this up? There should be a way to reference libraries without doing the directory change...see how NumPy etc. do this stuff

rahulpalamuttam:
Yea I'm waiting on #229 to be fixed and merged so I can rebase this.
The directory change is so libweld can find libweldrt - but with #229 it becomes trivial. I'll modify the pr so it loads in only libweld and remove the extra junk that were remnants of failed attempts.

python/setup.py
```diff
@@ -39,7 +39,7 @@ def has_ext_modules(self):
 
 setup(name='weld',
       version='0.0.1',
-      packages=['weld'],
+      packages=['weld', 'grizzly'],
```
wesm: Installing multiple top level packages might be a little bit counterintuitive for a lot of Python programmers. I would encourage you to set up a single top level namespace entry point (like `import weld.grizzly as gr`)

rahulpalamuttam:
All right so currently the packaging works with osx.
pip install --index-url https://test.pypi.org/simple/ pyweld
pip install --index-url https://test.pypi.org/simple/ grizzly

Note that the pip command being used cannot be the system pip.
You need to use a brew installed python or conda installed python.
The reason is that the system python uses an outdated openssl version which pypi rejects when pulling its packages. Manually upgrading openssl does not cause the system python to use the updated openssl.

build.rs
```diff
@@ -7,15 +7,15 @@ fn main() {
...
+        .arg(format!("{}/python/grizzly/grizzly", project_dir))
         .status()
         .unwrap();
     assert!(status.success());
 
     let status = Command::new("make")
         .arg("convertor")
         .arg("-C")
-        .arg(format!("{}/python/grizzly/", project_dir))
+        .arg(format!("{}/python/grizzly/grizzly", project_dir))
```
rahulpalamuttam: We shouldn't be making the convertor dylib as part of the weld installation step. Also since we're not invoking setup.py the Makefile will use the default "which python" function which may be different than the python you run "python setup.py install". I vote to just delete these lines.

python/grizzly/setup.py
```diff
@@ -0,0 +1,42 @@
...
+elif system == 'Darwin':
+    numpy_convertor = "numpy_weld_convertor.dylib"
+else:
+    raise OSError("Unsupported platform {}", system)
+
+class Install(install):
+    def run(self):
+        install.run(self)
+        python_executable = sys.executable
+        protoc_command = ["make -C " + self.install_lib + "grizzly/ EXEC=" + python_executable]
```
rahulpalamuttam: Should we invoke make clean before this?

deepakn94: Yeah, that sounds like a good idea

build.rs
deepakn94: Agreed, let's remove this -- `setup.py` calls `make convertor` now, right?

python/grizzly/setup.py
```diff
@@ -0,0 +1,42 @@
+import os
```
deepakn94: Should we rename this to `pygrizzly` or something, since `grizzly` is taken on PyPI?

PR #227, opened 2017-09-10
Only initialize weld runtime once

hvanhovell:
Weld runtime is initialized when a weld module is compiled. This initialization overwrite the current runtime state. This is problematic when you try to run multiple different modules at the same time. This PR adds a guard to `weld_runtime_init` to make sure it is only initialized once.


hvanhovell:
cc @mateiz @sppalkia

sppalkia:
Thanks!

PR #228, opened 2017-09-10
Iterate over empty vectors

hvanhovell:
It is currently impossible to iterate over empty vectors. If you try this weld will throw an `BadIteratorLength` error.

This is caused by the logic used to check iterator lengths. This calculates the end position of the iterator by applying the following steps in pseudo code:
```
val strides = numIterations - 1 // This yields -1 for an empty iterator
val delta = strides * stride // This yields -1 for an empty iterator and the normal stride size of 1
val end = start + delta // This yields -1 for an empty iterator, the normal stride size of 1 and a start of 0
if (end > numElements) {
  // Kaboom
}
```
The problem is that the comparison uses an unsigned comparison operator. `-1` will be interpreted as `2^64` which is (most likely) bigger than any input collection. This PR fixes this by using a signed comparison.

### DIFF
--- a/tests/integration_tests.rs
+++ b/tests/integration_tests.rs
@@ -799,6 +799,20 @@ fn nested_if_statement_with_builders_loop() {
     unsafe { free_value_and_module(ret_value) };
 }
 
+fn empty_appender_loop() {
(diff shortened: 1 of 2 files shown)


hvanhovell:
cc @mateiz @sppalkia

PR #229, opened 2017-09-10
Make Runtime library statically linked

hvanhovell:
The current weld runtime library is dynamically loaded as a separate library. This is quite cumbersome when you use weld from java or python, such applications typically have less control over installing and loading native libraries.

This PR tries to address this by statically linking the runtime library. This was not done before because the symbols needed by the generated code are removed during linking (because they appear not to be). This was especially problematic for the executables produced, the repl and the integration tests. This problem is fixed in two ways:
- All the runtime functions (symbols) are exported in `runtime.rs` source file. This is needed to export the symbols in the library. This is very similar to approach take before https://github.com/weld-project/weld/commit/7a425988237f209d45649053d334403ae4dffd2d was merged.
- We add `-export_dynamic` to the link options, and we make sure that the static library only uses one object file. Both steps are needed to make integration tests and the repl working  again. The downside to this approach is that the executables (not the libraries) almost double in size because all symbols are exported. In the future we could try to put executables on a diet by passing an explicit exported symbol list to the linker.

I have also removed the `run_named` related functionality because this is not needed anymore.

The PR has been tested on Mac and Linux.


mateiz:
This is great. For some reason, when I tried to just export these functions in a Rust module, it didn't work -- they still weren't all in the final dynamic library. Maybe I just did something wrong. I didn't have the `#link` attribute there for example. Anyway, we should merge this.

PR #231, opened 2017-09-11
Fix the error message when NumPy is not installed.

sarutak:
When we try to build weld without NumPy installed, the error message may somewhat confuse us.
Let's make the error message more meaningful.

### DIFF
--- a/python/grizzly/Makefile
+++ b/python/grizzly/Makefile
@@ -2,8 +2,13 @@ OS=$(shell uname -s)
 LLVM_VERSION=$(shell llvm-config --version | cut -d . -f 1,2)
 
 PYTHON_HEADER_INCLUDE = $(shell python-config --includes)
-NUMPY_HEADER_INCLUDE = -I$(shell python -c "import numpy; print numpy.get_include()")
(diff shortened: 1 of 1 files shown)


python/grizzly/Makefile
```diff
@@ -2,8 +2,13 @@ OS=$(shell uname -s)
 LLVM_VERSION=$(shell llvm-config --version | cut -d . -f 1,2)
 
 PYTHON_HEADER_INCLUDE = $(shell python-config --includes)
-NUMPY_HEADER_INCLUDE = -I$(shell python -c "import numpy; print numpy.get_include()")
+NUMPY_HEADER_INCLUDE = -I$(shell python -c "import numpy; print numpy.get_include()" 2>/dev/null)
 PYTHON_LDFLAGS = $(shell python-config --ldflags)
+
+ifeq (${NUMPY_HEADER_INCLUDE}, -I)
+  $(error NumPy is needed to be installed)
```
sppalkia: Thanks for the contribution! can you change the message to:

"Error: NumPy installation not found"

sarutak: O.K. I'll modify it.

sppalkia:
Looks good, thanks!

PR #232, opened 2017-09-12
Add entries to .gitignore to ignore backup files of Emacs

sarutak:
Emacs has 3 types of backup files like vim's .swp file.
I think it's better to ignore those files.


sppalkia:
Thanks!

PR #240, opened 2017-09-27
Build refactor

rahulpalamuttam:
1. Removes the make commands from build.rs that were used to build the convertor dylib.
2. Change the package name grizzly to pygrizzly
3. Added a binary extension for libweld in pyweld/setup.py which somehow allowed auditwheel to stop complaining about libweld.so being part of the python wheel. I could run auditwheel successfully and it changed the platform tag which allowed me to upload to pypi.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -7,22 +7,6 @@ fn main() {
     let status = Command::new("make")
         .arg("clean")
         .arg("-C")
-        .arg(format!("{}/python/grizzly/grizzly", project_dir))
-        .status()
-        .unwrap();
-    assert!(status.success());
-
-    let status = Command::new("make")
-        .arg("convertor")
-        .arg("-C")
-        .arg(format!("{}/python/grizzly/grizzly", project_dir))
-        .status()
-        .unwrap();
-    assert!(status.success());
-
-    let status = Command::new("make")
-        .arg("clean")
-        .arg("-C")
         .arg(format!("{}/weld_rt/cpp/", project_dir))
         .status()
         .unwrap();
--- a/python/README.md
+++ b/python/README.md
@@ -4,7 +4,7 @@ This directory contains the Weld Python API and the Grizzly implementation.
 
 ### Prerequisites
 
-Build and run tests for Weld (the instructions for this are in the main `README.md`).
+Build and run tests for Weld (the instructions for this are in the main `README.md`. <br />
 Make sure the `WELD_HOME` environment variable is set as detailed in the main
 `README.md`.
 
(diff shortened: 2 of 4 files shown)


rahulpalamuttam:
This is the general procedure for packaging pyweld and grizzly.

On OSX:
pyweld -
$ python setup.py bdist_wheel
Then upload the wheel found under dist/

On Linux:
$python setup.py bdist_wheel
$ auditwheel repair dist/*.whl
Then upload the wheel found under dist

With Grizzly, we do not want to build the convertor dylib, since we need it to link to the appropriate libpython.a. So we need to build the convertor as a post install step. For this reason, we package grizzly as a source distribution.

$python setup.py sdist
Then upload the tar.gz file.

wesm:
Are you using the manylinux1 toolchain?

wesm:
(for Linux). On OS X there are some related scripts (see multibuild project used here https://github.com/apache/arrow-dist/tree/master/python-macos)

rahulpalamuttam:
@wesm I did explore the manylinux1 toolchain. I may have misunderstood some of the steps, but I ran into some problems trying to build the project on the centos 5 docker containers. Namely, when trying to yum install certain tools (like llvm) I would see "Reducing CentOS-5 - libselinux to included packages only". I guess I can still build the libweld.so on the host linux machine, and then copy that to the docker containers when manylinux1 builds its wheels. This would give us an automated way of building manylinux wheels. I'm just doing it manually now via auditwheel.

xhochy:
@rahulpalamuttam Two things you need to take care:

* `manylinux1` wheels should be build inside of the docker container using the provided gcc, don't simply depend on auditwheel to do the checks.
* You should never link against `libpython` only if you really need to instantiate a new interpreter. If you need to do this you're out of luck for providing reproducible wheels. Especially as the interpreter API has breaking changes in patch releases of Python (<3.6).

rahulpalamuttam:
@xhochy I agree, however, we need clang.

We are not packing grizzly in a wheel. It's a source distribution so I'm only uploading the tar.gz.
We should get rid of the libpython dependency.

rahulpalamuttam:
@sppalkia addressed

sppalkia:
Is this ready to be merged? @rahulpalamuttam

sppalkia:
Actually Rahul, sorry, we also just merged Pari's numpy thing; can you pull that in and make sure the installation for that is the same? Or just make sure it works; doesn't need to be pip-installable necessarily.

rahulpalamuttam:
@sppalkia It's ready to be merged.
There are 4 numpy tests that are failing - but those are meant to fail.
The other 40 tests pass with the installed bindings.

PR #242, opened 2017-09-28
Performance graph produced from each Travis run

deepakn94:

PR #243, opened 2017-10-03
Add Python Files to Gitignore

sppalkia:


deepakn94:
Looks ok to me -- what did you run to verify this worked?
Also, tests are failing, but that seems to be my fault -- should work after #242 is merged.

python/pyweld/setup.py
```diff
@@ -30,6 +30,8 @@ def move_file(self, filename, directory):
         source = filename
         dir, name = os.path.split(source)        
         destination = os.path.join(self.build_lib + "/" + directory + "/", name)
+        makedir = self.build_lib + "/" + directory
+        os.system("mkdir -p {}".format(makedir))
```
rahulpalamuttam: The build directory should be created automatically on setup.py install.
I'm guessing you're doing this because setup.py develop is failing?
The PR #240 fixes setup.py develop.

sppalkia: Ah okay, I'll revert this then.

Issue #244, opened 2017-10-04
Iterate can cause Stack Frame Exhaustion


sppalkia:
Currently, iterate seems to use recursion to move through basic blocks. Large numbers of iterations can thus cause a segmentation fault because stack space is exhausted.

sppalkia:
resolved by #245

PR #245, opened 2017-10-06
Fix for Iterate

sppalkia:
This patch fixes a bug where `Iterate` expressions with nested `For` expressions could exhaust the stack by iterating via recursion. Specifically, the issue occurs if `For` loops are small enough that the runtime is not invoked and the loop body/continuation is called as function call.

This patch circumvents the problem by adding an "parallelize" annotation along with a transformation which forces calling into the parallel runtime for all for loops within an iterate. This could harm performance, so it needs a long term fix.


weld/sir.rs
```diff
@@ -654,15 +656,15 @@ fn sir_param_correction(prog: &mut SirProgram) -> WeldResult<()> {
 }
 
 /// Convert an AST to a SIR program. Symbols must be unique in expr.
-pub fn ast_to_sir(expr: &TypedExpr) -> WeldResult<SirProgram> {
+pub fn ast_to_sir(expr: &TypedExpr, multithreaded: bool) -> WeldResult<SirProgram> {
```
sppalkia: Added this to make it easier to do single vs. multithreaded optimizations in the SIR later.

sppalkia:
@hvanhovell this should fix the seg fault in the wide aggregation workload. Addresses #244

weld/llvm.rs
```diff
@@ -526,8 +533,17 @@ impl LlvmGenerator {
...
         let mut grain_size = DEFAULT_GRAIN_SIZE;
         if par_for.innermost {
-            ctx.code.add(format!("{} = icmp ule i64 {}, {}", bound_cmp, num_iters_str, grain_size));
-            ctx.code.add(format!("br i1 {}, label %for.ser, label %for.par", bound_cmp));
+            // Determine whether to always call parallel, always call serial, or
+            // choose based on the loop's size.
+            if par_for.parallelize {
+                ctx.code.add(format!("br label %for.par"));
+            }
+            else if self.multithreaded {
```
jjthomas: move this up a line so that it's on the same line as the previous closing curly brace

jjthomas: to make this code clearer i would actually have
`if par_for.parallelize {
...
} else {
  if (self.multithreaded) {
  ...
  } else {
  ...
  }
}`

weld/ast.rs
```diff
@@ -276,6 +278,14 @@ impl Annotations {
         self.vectorize = Some(vectorize)
     }
 
+    pub fn parallelize(&self) -> &Option<bool> {
```
jjthomas: i think `parallelize` is a misleading for this ... it should be something like `always_call_runtime` or `always_use_runtime`

sppalkia:
Good point — do you think multithreaded should also be its own config?
Right now it just checks the number of threads set, but that’s slightly
different, since this is specifying how to compile the module (not how to
run it)

jjthomas:
yeah setting the thread count and compile time and runtime is very confusing ... i think we should split up the conf into a CompileConf and RunConf, and thread count should only be in the RunConf. CompileConf can have something like a `compile serial` flag, which would emit just serial code, and then you could have a check at the start of the execution to make sure that the number of threads is always 1.

sppalkia:
@jjthomas addressed your comments - yeah, might make more sense to have to Confs. For now I just made a separate option, I can break them up into compile/run in a different PR. We should also give the configuration keys consistent names.

jjthomas:
looks good, feel free to merge

PR #248, opened 2017-10-10
Lazy Helpers

sppalkia:
Lazily generates helper functions (`cmp`, `eq`, and `hash`) for complex/nested types. This marginally improves compile time and drastically reduces the LLVM code size for larger programs.


jjthomas:
didn't look too closely but seems fine

PR #252, opened 2017-10-18
Birth analysis

rahulpalamuttam:
Still a work in progress.

Currently what works is:
This first PR has:
1. GroupBy Sums that can utilize mutliple columns for both keys and values
2. GroupBy and slice on each aggregated value vector

GroupBy sort and the pivot table are WIP


weld/type_inference.rs
```diff
@@ -828,6 +828,14 @@ fn push_type(dest: &mut PartialType, src: &PartialType, context: &str) -> WeldRe
                             for ty in tys {
                                 match *ty {
                                     Scalar(_) => {}
+                                    Vector(ref sty) => {
```
rahulpalamuttam: This is to enable merging in vec[i8]'s into groupmergers.

weld/type_inference.rs
```diff
@@ -167,9 +167,9 @@ fn infer_locally(expr: &mut PartialExpr, env: &mut TypeMap) -> WeldResult<bool>
                                 .zip(vec![key_type.clone(), value_type.clone()].iter_mut()) {
                             changed |= try!(push_type(field_ty, child_expr_field, "ToVec"));
                         }
-                    } else {
-                        return weld_err!("Internal error: toVec argument needs to be a dict");
-                    }
+                    }//  else {
```
rahulpalamuttam: The type inference fails for the following kind of expression
`let rs = result(for(x, {dictmerger[i32,i32,+], dictmerger[i64,i64,+]}, |b,i,e| {b.$0, b.$1}))
  let agg_dict = result(rs.$0)
`

Here agg_dict has an "Unkown" type and so we can't call tovec on it. 
I had to comment this else statement out so I could proceed with generating the code.
There's probably a more elegant solution.

python/grizzly/grizzly/grizzly.py
```diff
@@ -5,6 +5,139 @@
 from lazy_op import LazyOpResult, to_weld_type
 from weld.weldobject import *
 
+class DataFrameWeldExpr:
+    def __init__(self, expr, column_names, weld_type, is_pivot=False):
```
rahulpalamuttam: I needed to introduce this class to be able to operate on dataframes that were tied to code that basically zipped their columns. I.e. some sort of zip expression.

Introducing another class just to do this is probably not the best option and we should find a way to merge the code here back into the original DataFrameWeld class.

python/grizzly/grizzly/grizzly.py
```diff
@@ -116,21 +250,10 @@ def __setitem__(self, key, value):
                 key
             )
 
-    def __getattr__(self, key):
```
rahulpalamuttam: This function needs to be deleted. Otherwise, we cannot access any of the member variables of the class.

weld/type_inference.rs
rahulpalamuttam: This might actually fixed with shoumik's recent change fixing type inference on lets.

weld/type_inference.rs
```diff
@@ -338,6 +338,10 @@ fn infer_locally(expr: &mut PartialExpr, env: &mut TypeMap) -> WeldResult<bool>
                 }
                 changed |= try!(push_type(&mut expr.ty, &data.ty, "Sort"));
                 Ok(changed)
+            } else if data.ty == Unknown {
+                // TODO : The type inferencing is giving unkowns for
```
rahulpalamuttam: This is a hack for now. 
For some reason, it's not inferring the type of a let assigned to a vector of structs.

rahulpalamuttam: A @sppalkia 
Should I get rid of this line as well now with the type inference changes?

rahulpalamuttam:
Never seen this before.
Apparently, grizzly_test.py is failing with illegal instruction error.

python/grizzly/grizzly/numpy_weld_convertor.cpp
```diff
@@ -202,58 +202,83 @@ struct numpy_to_weld_char_arr_arr_args {
...
+//     return t;
+//   }
+
+//   // #pragma omp parallel for // uncomment to enable parallel encode
+//   for (int i = 0; i < num_threads; i++) {
+//     numpy_to_weld_char_arr_arr_helper(&args[i]);
+//   }
+
+//   return t;
+// }
```
rahulpalamuttam: For some reason the uncommented region fails on linux.

examples/python/grizzly/birth_analysis_grizzly.py
```diff
@@ -0,0 +1,53 @@
+import pandas as pd
+import grizzly.grizzly as gr
+import numpy as np
+import time
+import sys
+
+passes = ["loop-fusion", "infer-size", "short-circuit-booleans", "predicate", "vectorize", "fix-iterate"]
+years = range(1880, 2011)
+pieces = []
+#columns = ['year', 'name', 'sex', 'births']
```
deepakn94: Remove dead code

.travis.yml
```diff
@@ -1,6 +1,7 @@
 language: rust
 sudo: required
 dist: trusty
+group: deprecated-2017Q4
```
deepakn94: Can you put a comment here about why this is needed? (is it even needed after the convertor fix?)

examples/python/grizzly/birth_analysis_grizzly.py
```diff
@@ -0,0 +1,53 @@
...
+names = pd.concat(pieces, ignore_index=True)
+print "Size of names: %d" % len(names)
+
+def get_top1000(group):
+    # Note that there is a slight difference that arises
+    # with the name 'Leslyn', year '25' missing as the ordering
+    # in pandas for rows witht he same 'sort' value changes.
+    return group.sort_values(by='births', ascending=False).slice(0, 1000)
+
+#Time preprocessing step
```
deepakn94: Nit: Put a space here between `#` and `T`

examples/python/grizzly/movielens.py
```diff
@@ -0,0 +1,41 @@
...
+                        names=rnames)
+
+mnames = ['movie_id', 'title', 'genres']
+movies = pd.read_table('data/ml-1m/movies.dat', sep='::', header=None,
+                       names=mnames)
+
+start = time.time()
+data = pd.merge(pd.merge(ratings, users), movies)
+end = time.time()
+print data
```
deepakn94: Remove

examples/python/grizzly/scripts/replicate-csv
```diff
@@ -4,6 +4,32 @@ import argparse
...
+                continue
+            line_ele_list = [str(line_ele) for line_ele in line]
+
+            for factor in range(0, replicationFactor):
+                count += 1
+                new_year = years[count % len(years)]
+                # This is a hack. We need to get to the bottom of pandas ordering on sorts
+                # which is causing grizzly to include "Leslyn" when pandas does not
+                # for birth analysis (sort on groupmerger).
+                num_diff = str(int(line_ele_list[2]) + count)
```
deepakn94: I don't fully understand the reason for this hack

python/grizzly/grizzly/grizzly.py
```diff
@@ -5,6 +5,367 @@
 from lazy_op import LazyOpResult, to_weld_type
 from weld.weldobject import *
 
+class DataFrameWeldLoc:
```
deepakn94: Comment on what this is used for would be good

python/grizzly/grizzly/grizzly.py
```diff
@@ -5,6 +5,367 @@
...
+                        index_expr_predicate
+                    ),
+                    self.df.column_names,
+                    self.df.weld_type,
+                    is_pivot=True
+                )
+            # TODO : Need to implement for non-pivot tables
+        raise Exception("Cannot invoke getitem on an object that is not SeriesWeld")
+
+class WeldLocIndexer:
```
deepakn94: Same comment as above

python/grizzly/grizzly/grizzly.py
```diff
@@ -5,6 +5,367 @@
...
+                        series.df,
+                        series.column_name,
+                        series.index_type,
+                        series.index_name
+                    )
+
+            # TODO : Need to implement for non-pivot tables
+        raise Exception("Cannot invoke getitem on non SeriesWeld object")
+
+class DataFrameWeldExpr:
```
deepakn94: Why do we have so many of these classes

python/grizzly/grizzly/grizzly_impl.py
```diff
@@ -213,6 +332,139 @@ def element_wise_op(array, other, op, ty):
                                           "ty": ty, "op": op}
     return weld_obj
 
+def unzip_columns(expr, column_types):
+    """
+    Zip together multiple columns.
```
deepakn94: Fix comment

python/grizzly/grizzly/grizzly_impl.py
```diff
@@ -213,6 +332,139 @@ def element_wise_op(array, other, op, ty):
                                           "ty": ty, "op": op}
     return weld_obj
 
+def unzip_columns(expr, column_types):
```
deepakn94: What is the usecase of this method?

examples/python/grizzly/scripts/replicate-csv
rahulpalamuttam: When we do the groupby_sort, we are sorting by number of births. However, some names have the same number of births associated with it. We can't seem to match whatever ordering pandas is using when it places items with the same sort_key value. This causes taking topk in a sorted vector to be off.

rahulpalamuttam: I think we should get rid of this snippet and just use the transformCSV function that's already there. We should figure out what's going on with the sort instead of changing the actual values of the dataset.

python/grizzly/grizzly/grizzly.py
rahulpalamuttam: I found it difficult dealing with DataFrameWeld which kept track of un-materialized columns and materialized columns. Instead treating dataframes as a zip of columns from the onset makes it very easy to deal with. We should really just merge these two classes.

python/grizzly/grizzly/grizzly_impl.py
rahulpalamuttam: Pulling apart a zipped expression.
This is mostly useful when materializing the columns of a dataframe. 
We can't materialize a vector of structs, and convert that to numpy arrays. So we unzip to have a struct of vectors which we can then convert to numpy arrays.

examples/python/grizzly/birth_analysis_grizzly.py
rahulpalamuttam: Adressed.

.travis.yml
rahulpalamuttam: Addressed. It's not necessary.

examples/python/grizzly/birth_analysis_grizzly.py
rahulpalamuttam: Addressed.

examples/python/grizzly/movielens.py
rahulpalamuttam: Addressed.

python/grizzly/grizzly/grizzly.py
rahulpalamuttam: Addressed.

python/grizzly/grizzly/grizzly.py
rahulpalamuttam: Addressed.
We can probably merge the WeldLocIndexer and DataFrameWeldLoc classes. 
LocIndexer and DataFrameLoc were two types of objects I encountered in Pandas but they seem to be having the same functionality which is just selecting rows based on an index column.

python/grizzly/grizzly/grizzly_impl.py
rahulpalamuttam: Addressed

examples/python/grizzly/birth_analysis_grizzly.py
```diff
@@ -0,0 +1,52 @@
...
+top1000 = top1000.evaluate(True, passes=passes).to_pandas()
+end0 = time.time()
+
+start1 = time.time()
+top1000 = gr.DataFrameWeld(top1000)
+top1000names = top1000['name']
+all_names = top1000names.unique()
+lesley_like = all_names.filter(all_names.lower().contains('lesl'))
+
+filtered = top1000.filter(top1000names.isin(lesley_like))
```
sppalkia: eval here for -CLO

examples/python/grizzly/movielens_grizzly.py
```diff
@@ -0,0 +1,49 @@
...
+print data
+
+start1 = time.time()
+data = gr.DataFrameWeld(data)
+mean_ratings = data.pivot_table('rating', index='title', columns='gender',
+                                aggfunc='mean')
+
+
+ratings_by_title = data.groupby('title').size()
+active_titles = ratings_by_title.index[ratings_by_title >= 250]
```
sppalkia: -CLO evaluate

examples/python/grizzly/movielens_grizzly.py
```diff
@@ -0,0 +1,49 @@
...
+mean_ratings = mean_ratings.loc[active_titles]
+mean_ratings['diff'] = mean_ratings['M'] - mean_ratings['F']
```
sppalkia: -CLO evaluate

weld/llvm.rs
```diff
@@ -3453,47 +3458,84 @@ impl LlvmGenerator {
...
+                                key_ll_ty = key_ll_tyc.replace("%", "");
+                                key_ll_sym = key_ll_symc;
+                                // Generate prelude code
+                                let keyfunc_ctx = &mut FunctionContext::new(false);
+                                for (arg, ty) in keyfunc.locals.iter() {
+                                    let arg_str = llvm_symbol(&arg);
+                                    let ty_str = self.llvm_type(&ty)?;
+                                    keyfunc_ctx.add_alloca(&arg_str, &ty_str)?;
+                                }
+                                for b in keyfunc.blocks.iter() {
```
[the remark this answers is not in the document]
rahulpalamuttam: Just handle terminators that are branches and jmpblocks.
Generate those terminators.
If it's a program return skip, anything else throw an error.
Avoid generating terminators for the last block.

weld/type_inference.rs
```diff
@@ -821,6 +821,14 @@ fn push_type(dest: &mut PartialType, src: &PartialType, context: &str) -> WeldRe
                             for ty in tys {
                                 match *ty {
                                     Scalar(_) => {}
+                                    Vector(ref sty) => {
+                                        match **sty {
+                                            Scalar(_) => {}
+                                            _ => {
+                                                return weld_err!("Commutative merge builders can only contain scalar vectors in structs");
```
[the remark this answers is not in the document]
rahulpalamuttam: How would we merge in a something of the following type {vec[{i8, i8}], vec[{i8, i8}]}?
Or even {vec[vec[i8]], vec[vec[i8]]}

weld/llvm.rs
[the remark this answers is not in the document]
rahulpalamuttam: addressed

rahulpalamuttam:
@jjthomas I addressed the comments

weld/llvm.rs
```diff
@@ -3453,84 +3453,75 @@ impl LlvmGenerator {
...
+                                                keyfunc_ctx.code.add(format!("br i1 {}, label %b.b{}, label %b.b{}", cond_tmp, on_true, on_false));
+                                            } else if let Terminator::JumpBlock(block) = b.terminator {
+                                                keyfunc_ctx.code.add(format!("; {}", b.terminator));
+                                                if self.trace_run {
+                                                    self.gen_puts(&format!("  {}", b.terminator), ctx);
+                                                }
+                                                keyfunc_ctx.code.add(format!("br label %b.b{}", block));
+                                            } else if let Terminator::ProgramReturn(_) = b.terminator {
+                                            } else {
+                                                return weld_err!("Can't have intermediate return in sort key function");
```
jjthomas: change this message to "Can't have terminator other than Branch or JumpBlock in sort key function"

rahulpalamuttam: addressed.

PR #254, opened 2017-10-19
Reduce small loop overhead

jjthomas:
1) increase outer loop grain size to 4096
2) eliminate bounds checks for simple single-iter (no explicit start and end) cases
3) minimize thread ID retrievals
4) local stack-based mergers with lazy creation of global mergers

performance on a worst-case loop like this

```
merger[+] m
for v in vs {
  merger[+] n
  for e in v { // only one iteration
    merge(n, e)
  }
  merge(m, result(n))
}
result(m)
```

is still 4x off the C version ... used to be 500x though


jjthomas:
Performance numbers seem fine:

```
++++++++++++++++++++++++++++++++++++++
tpch_q1
++++++++++++++++++++++++++++++++++++++
p=1.0, n=100000000
Weld: 0.9762 +/- 0.0145 seconds
Weld compile time: 0.1506 +/- 0.0025 seconds
Single-threaded C++: 0.6325 +/- 0.0040 seconds


++++++++++++++++++++++++++++++++++++++
tpch_q6
++++++++++++++++++++++++++++++++++++++
p=1.0, n=200000000
Weld: 0.4015 +/- 0.0015 seconds
Weld compile time: 0.1045 +/- 0.0010 seconds
Single-threaded C++: 0.4444 +/- 0.0013 seconds


++++++++++++++++++++++++++++++++++++++
vector_sum
++++++++++++++++++++++++++++++++++++++
n=400000000
Weld: 0.1627 +/- 0.0235 seconds
Weld compile time: 0.0993 +/- 0.0012 seconds
Single-threaded C++: 0.1536 +/- 0.0202 seconds


++++++++++++++++++++++++++++++++++++++
map_reduce
++++++++++++++++++++++++++++++++++++++
n=400000000
Weld: 0.1592 +/- 0.0207 seconds
Weld compile time: 0.0995 +/- 0.0016 seconds
Single-threaded C++: 0.1450 +/- 0.0011 seconds


++++++++++++++++++++++++++++++++++++++
data_cleaning
++++++++++++++++++++++++++++++++++++++
s=10, f=data/311-service-requests-sf=%d.csv
Python->Weld: 0.0105 +/- 0.0011 seconds
Weld: 0.0521 +/- 0.0013 seconds
Grizzly: 0.3132 +/- 0.0045 seconds
Weld->Python: 0.0001 +/- 0.0000 seconds
Weld compile time: 0.2176 +/- 0.0019 seconds
Pandas: 0.2801 +/- 0.0023 seconds


++++++++++++++++++++++++++++++++++++++
crime_index
++++++++++++++++++++++++++++++++++++++
s=10, f=data/us_cities_states_counties_sf=%d.csv
Python->Weld: 0.0181 +/- 0.0028 seconds
Weld: 0.0395 +/- 0.0015 seconds
Grizzly: 0.5676 +/- 0.0035 seconds
Weld->Python: 0.0001 +/- 0.0000 seconds
Weld compile time: 0.4407 +/- 0.0032 seconds
Pandas: 0.0760 +/- 0.0023 seconds


++++++++++++++++++++++++++++++++++++++
crime_index_simplified
++++++++++++++++++++++++++++++++++++++
s=10, f=data/us_cities_states_counties_sf=%d.csv
Python->Weld: 0.0090 +/- 0.0005 seconds
Weld: 0.0168 +/- 0.0001 seconds
Grizzly: 0.3604 +/- 0.0027 seconds
Weld->Python: 0.0000 +/- 0.0000 seconds
Weld compile time: 0.2905 +/- 0.0022 seconds
Pandas: 0.1612 +/- 0.0081 seconds

mateiz:
Hey James, I don't completely understand the times above -- is Grizzly still slower than Pandas? Why is that? I don't understand what the Grizzly time is in there versus the Weld time. Is it because Grizzly doesn't generate the optimal code yet, so the Weld number is for some other code?

Also, please run these at 10-100x larger scale factor so we can tell more about the differences. All of the times are very small now.

wesm:
Aside: as we start delivering the Arrow-based pandas2 runtime over the course of 2018 I will try to post Arrow/pandas2-based versions of these benchmarks so you have another data point to compare against. pandas has very poor data locality on non-numeric data and so the biggest speed improvements will be on strings and non-number columns. 

Multicore scheduling and elimination of temporaries will also yield major boosts. At some point in 2018 I would like to see how to enable Weld to run atop Arrow buffer layouts so that we can dynamically generate new operator kernels and inject them into the pandas2 runtime.

jjthomas:
I think the Grizzly time just includes the compilation time. That should be the only difference. @deepakn94 can you confirm

mateiz:
OK, in any case, try running it on even 100x more data to see what it does.

deepakn94:
Yes, Grizzly includes both compute time + compilation time.
I need to improve time logging for these Pandas workloads...

PR #256, opened 2017-10-20
Change Exception to AttributeError in Grizzly wrappers

### LINKED ISSUE #255: Change Exception to AttributeError in __getattr__ handlers of Grizzly dataframe and series wrappers
AFAIK most interactive environments (read: ipython notebooks) make use of `dir()` special function in order to provide autocomplete suggestions. Be default, `dir()` tries to call the `__dir__()` special method if it can, falling back to a default implementation if an `AttributeError` is raised. Most of the weld wrappers seem to implement a custom `__getattr__()` method which raises an exception when unrecognized attributes are provided as keys, and right now `__dir__()` is considered an unrecognized attribute. The provided `__getattr__()` raises `Exception`, and since `dir()` doesn't catch `Exception`, this unfortunately causes basic autocomplete functionality to break.

Generally speaking, other libraries may expect `AttributeError` to be raised for missing attributes.

A straightforward way to fix this is to simply raise `AttributeError` instead of Exception in these `__getattr__()` handlers.

smacke:
In DataFrameWeld and SeriesWeld, there are custom __getattr__ handlers
implemented that raise Exception if asked for an unsupported key. Since
these do not implement the __dir__ handler, this causes default dir()
functionality to break, since the behavior of dir() is to first ask for
the __dir__ attribute if it exists, and then fall back to some default
behavior if it catches an AttributeError. Since it only catches
AttributeError and not Exception, dir() no longer works, notably
preventing autocomplete for interactive environments like ipython
notebooks. We provide the straightforward fix in this commit by simply
raising AttributeError instead of Exception in the implementation of
__getattr__ for these classes.

Note that things that are retrieved through __getattr__, such as
'values' for SeriesWeld, will still not work with autocomplete, since
the default dir() implementation is unaware of things that are retrieved
by falling back to __getattr__.

This commit fixes #255

### DIFF
--- a/python/grizzly/grizzly/grizzly.py
+++ b/python/grizzly/grizzly/grizzly.py
@@ -127,7 +127,7 @@ class DataFrameWeld:
                     weld_type,
                     dim
                 )
-        raise Exception("Attr %s does not exist" % key)
+        raise AttributeError("Attr %s does not exist" % key)
 
     def _get_column_names(self):
         """Summary
@@ -311,7 +311,7 @@ class SeriesWeld(LazyOpResult):
                 self.df,
                 self.column_name
             )
-        raise Exception("Attr %s does not exist" % key)
+        raise AttributeError("Attr %s does not exist" % key)
 
     def unique(self):
         """Summary


sppalkia:
Very interesting -- thanks so much for the fix!

PR #257, opened 2017-10-24
Weld Transform Compilation Improvements

sppalkia:
This PR adds a number of improvements, in particular to compilation time reporting and overall compilation performance for some of the transforms:

* Adds some logging information to report how long Weld compilation takes (e.g., each pass, parsing, type inference, LLVM code generation, etc.)

* Adds a new `annotations` module and reworks the annotation code to make the Annotation struct smaller. This improves hashing performance for types.

* Adds a new `expr_hash` module to hash expressions. This is a fast way to compare whether two expressions are equal.

* Rework`uniquify` to be much faster (it no longer clones sub-expressions)

* Rework the pass manager to be faster (uses the aforementioned `expr_hash` module to determine whether a pass should continue

### Random Fixes and Improvements

* Fix colored output for logging (used to all be the same color)
* Add a verbosity command line option for the REPL, so the log-level is configurable:

   ```
   repl [-v none|error|warn|info|debug|trace]
   ```
   The default log level is `debug`.


weld/expr_hash.rs
```diff
@@ -0,0 +1,189 @@
...
+                    Entry::Occupied(ref ent) => {
+                        ent.get().hash(&mut self.hasher);
+                    },
+                    _ => {
+                        return weld_err!("Undefined symbol {}", sym);
+                    }
+                }
+            }
+            BinOp {ref kind, .. } => {
+                kind.hash(&mut self.hasher);
```
jjthomas: why don't you call `from_expr` on the children of the `BinOp`?

sppalkia: At the bottom of this match there's a `for child in expr.children()` that handles the subexpressions - the match part just handles the non-subexpression fields (e.g., the `BinOpKind` or the actual literal value).

sppalkia:
@jjthomas added an explanation on this -- its mostly useful because it avoids having to clone the whole AST before mutating it, not because its faster than the actual comparison.

PR #260, opened 2017-10-25
Debug options to write Weld, SIR, and LLVM code to a file

sppalkia:
This adds the `weld.compile.dumpCode` and `weld.compile.dumpCodeDir` options to the weld configuration. The former causes Weld, SIR, and LLVM code to be dumped to separate files, and the latter specifies the directory to write the files to.

In the REPL, these options can be set using `setconf`, or can be set using the command line options `-D <logdir>` and the flag `-d` (to enable code dumping).


jjthomas:
looks good

PR #261, opened 2017-10-27
Implement simple sort

rahulpalamuttam:
Currently works with vectors of scalar type.
I'm still running into a few issues with implementing the sort function with a general typed key function.

We clone the vector before sorting and returning it.

I could use some pointers on how to check for dependencies on vectors so we can optionally generate the code to clone the vector before sorting. This wasn't done in the scala version, and I didn't completely figure out how to do it here.


rahulpalamuttam:
All right, here's a working sort with the key function.
I've made some comments along the way.
The thing to look at is how we generate the keyfunction code.

Right now, I make an entirely new SirFunction for it and actually pass the SirFunction as a parameter of in the Sort statement. 

I tried going down the track of creating a Lambda Statement and appending it to the current function block but the eventual result was a function implemented inside of a function (among other rust typing issues that came up). 

The constraints imposed on sort are
1. The sort return type is the same as its input type which is a vec.
2. The key function must only return scalars. Also the return type of the key function cannot be a reference to the return value. 
(returning stack references which get cleaned up once we return from the function results in nothing when we load from that address).
3. The parameter type of the keyfunction must be the element type of the vector we are sorting.
4. The keyfunction lives in its own context (no shared variables with the context that sort was called from).

weld/sir.rs
```diff
@@ -49,6 +49,11 @@ pub enum Statement {
         index: Symbol,
         size: Symbol,
     },
+    Sort {
+        output: Symbol,
+        child: Symbol,
+        keyfunc: SirFunction,
```
rahulpalamuttam: Thoughts on this approach?
I didn't find a way to make a Lambda symbol and using it properly.
There probably is a solution where having a Lambda symbol here is possible.

weld/llvm.rs
```diff
@@ -2384,6 +2384,96 @@ impl LlvmGenerator {
...
+                            keyfunc_ctx.add_alloca(&arg_str, &ty_str)?;
+                        }
+                        for b in keyfunc.blocks.iter() {
+                            for s in b.statements.iter() {
+                                self.gen_statement(s, keyfunc, keyfunc_ctx)?
+                            }
+                        }
+
+                        // Add key function and sort prelude code
+                        let mut func_gen = IdGenerator::new("%keyfunc");
```
jjthomas: you need a single global function generator, otherwise you'll get duplicated names

weld/llvm.rs
```diff
@@ -2384,6 +2384,96 @@ impl LlvmGenerator {
...
+                            "define {} {}({}) {{",
+                            key_ll_ty,
+                            func_str,
+                            str_args));
+                        self.prelude_code.add(keyfunc_ctx.alloca_code.result());
+                        self.prelude_code.add(keyfunc_ctx.code.result());
+                        self.prelude_code.add(format!("{}.ret = load {}, {}* {}", key_ll_sym, key_ll_ty, key_ll_ty, key_ll_sym));
+                        self.prelude_code.add(format!("ret {} {}.ret\n}}", key_ll_ty, key_ll_sym));
+                        let name = self.vec_names.get(elem_ty).unwrap();
+                        self.prelude_code.add(format!(
```
jjthomas: if you do two different sorts on the same vector type you'll add this prelude code twice, which will cause an name conflict error

weld/sir.rs
jjthomas: I think this is fine, I think the function will get inlined so there shouldn't be  performance issue ... you should run performance numbers though to confirm

weld/llvm.rs
rahulpalamuttam: Good call. I appended the key-function name {FUNC} as part of the sort function name in vector_sort.ll. So there can exist sort functions with the same type signatures, but they will differ since the key-function will be different. This also causes duplicated code being generated in cases where we call sort with the same key functions, but that should be a minor issue, right? I'm guessing that happens anyway due to function inlining.

jjthomas:
ok looks good, yeah duplication is not a big issue ... if you could just do a small performance test (maybe just sorting like 100m integers) and report the times in Weld and C, that would be good

rahulpalamuttam:
All right here are the results of the comparison.
https://docs.google.com/spreadsheets/d/13X1T8vn-eAW-c9dumG3Vtn21Mo6_kl42J0cxgorY8F8/edit#gid=0

This a bit of apples to oranges comparison since I used subtraction for the comparison function in the c code and we generate more complicated llvm code for the comparison involving the equality operator, branch, and less than operator. Using subtraction for i32's and i64's as the result of the comparison operator works for sorting and does result in a slightly faster sort. However, this breaks when we are comparing vectors (the vector comparison test fails). 

There is another solution where we get rid of the branch and instead multiply the result by 0 or 1 for integer scalars. It seems to be slightly faster (>%5).

PR #262, opened 2017-10-28
C++ Header Generation Tooling

sppalkia:
This PR adds a tool to generate C++ headers for argument/returns types in a Weld program. As an example, if we have the file called `program.weld` with the program:

```
|s: {i32, f32}| [s, s]
```

Running `target/release/hdrgen -i program.weld` will output the following C++ header file:

```c++
#ifndef _WELD_CPP_HEADER_
#define _WELD_CPP_HEADER_

// THIS IS A GENERATED C++ FILE.

#include <stdint.h> // For explicitly sized integer types.
#include <stdlib.h> // For malloc

// Defines Weld's primitive numeric types.
typedef bool        i1;
typedef uint8_t     u8;
typedef int8_t      i8;
typedef uint16_t   u16;
typedef int16_t    i16;
typedef uint32_t   u32;
typedef int32_t    i32;
typedef uint64_t   u64;
typedef int64_t    i64;
typedef float      f32;
typedef double     f64;

// Defines the basic Vector struture using C++ templates.
template<typename T>
struct vec {
  T *ptr;
  i64 size;
};

template<typename T>
vec<T> new_vec(i64 size) {
  vec<T> t;
  t.ptr = (T *)malloc(size * sizeof(T));
  t.size = size;
  return t;
}

struct struct0 {
  i32 _0;
  f32 _1;
};

struct struct1 {
  struct0 _0;
};

// Aliases for argument and return types.
typedef struct1 input_type;
typedef vec<struct0> return_type;

#endif /* _WELD_CPP_HEADER_ */
```

This makes it easier to write programs in C/C++ to debug Weld programs.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,6 +34,12 @@ name = "repl"
 path = "weld/bin/repl.rs"
 bench = false
 
+
+[[bin]]
+name = "hdrgen"
+path = "weld/bin/hdrgen.rs"
+bench = false
+
 [[test]]
 name = "integration_tests"
 path = "tests/integration_tests.rs"
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ You can join the discussion on Weld on our [Google Group](https://groups.google.
       - [Building Weld](#building-weld)
   * [Documentation](#documentation)
   * [Grizzly (Pandas on Weld)](#grizzly)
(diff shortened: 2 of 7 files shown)


sppalkia:
Merging this

PR #263, opened 2017-10-28
Fix for Hashing Annotations

sppalkia:
This fixes a potentially non-determinsitic/incorrect hash function for annotations. In the old hash function, we iterated over a `HashMap` and hashed each key and value. However, this may hash the keys in a different order on subsequent calls. Sorting the keys would likely be too slow for a hash function. We now use a `BTreeMap` instead, which can use to derive a hash function. These both fix the hash function and do not regress compilation performance.

### DIFF
--- a/weld/annotations.rs
+++ b/weld/annotations.rs
@@ -1,10 +1,9 @@
 use std::fmt;
 
-use std::collections::HashMap;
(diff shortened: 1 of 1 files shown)


deepakn94:
LGTM, merging.

PR #264, opened 2017-10-29
Modularize Transforms

sppalkia:
This pull request reorganizes the transform code to be in different modules. The `transforms.rs` file was getting a bit too large -- this puts each class of transformations in a separate submodule under the `transforms` module.


deepakn94:
LGTM; merging.

PR #265, opened 2017-10-31
Nditer

parimarjan:
The main structural changes involves adding a new kind, NdIter, to IterKind:
    pub enum IterKind {
        ScalarIter, // A standard scalar iterator.
        SimdIter, // A vector iterator.
        FringeIter, // A fringe iterator, handling the fringe of a vector iter.
        NdIter,     // multi-dimensional nd-iter
    }

and a couple of fields relevant to NdIter in struct Iter, and struct ParallelForIter, e.g,:
    pub struct ParallelForIter {
        pub data: Symbol,
        pub start: Option<Symbol>,
        pub end: Option<Symbol>,
        pub stride: Option<Symbol>,
        pub kind: IterKind,
        // NdIter specific fields
        pub strides: Option<Symbol>,
        pub shapes: Option<Symbol>,
    }

The code follows the same path as for other IterKinds, and the main changes are in the
code to generate the index of the next element in llvm.rs (gen_loop_iteration_start, and
gen_loop_iteration_end). And changing the bounds/num_iterations based on the shapes parameter also in llvm.rs (gen_loop_bounds_check, and gen_num_iters_and_fringe_start).

Besides this, I added a struct in llvm.rs (not sure if this was the best place for this?):
    pub struct VecLLVMInfo {
        pub ty_str: String,
        pub arr_str: String,
        pub prefix: String,
        pub len_str: String,
        pub el_ty_str: String,
    }
which was useful for many of the llvm routines I had.

I have only added two tests for it (testing with zip, and a basic op (log)), but I did test
it further using the numpy API where it was more natural as I could compare different outputs
of non-contig arrays with results numpy produces. Here, I had a simple case that simulates a
non-contiguous array using a 1-d rust array.

One issue that still remains is how to set the size of the array correctly: right now it is being
done in transforms.rs --> infer_size using the formula: len = end-start / stride; Instead, the new
formula should be shapes[0]*shapes[1]*...*shapes[n-1], but I wasn't sure how to emit weld code for this
using exprs...in weld it should be something like:

    for(shapes, merger[i64, *], |b, i, e| merge(b, e));

And using exprs, as in infer_size, I thought it would be something like:
    let b = exprs::newbuilder_expr(Merger(Scalar(I64), BinOpKind::Multiply))?;
    let m = exprs::merge_expr(b, ???);
    exprs::for_expr(iters[0].data.clone(), b, m, false)?

but I wasn't sure how to do the merge_expr --> in weld code I guess we have access to element "e",
but how do we get it here?

If this is resolved, then we won't need to pass in "end", and "stride" to nditer. So far, I was
just using stride=1, and end = start + real_len, which is functionally correct, but adds too
many parameters to nditer...


weld/llvm.rs
```diff
@@ -806,53 +837,98 @@ impl LlvmGenerator {
...
+                    let data_llvm_info = self.get_array_llvm_info(func, ctx, &iter.data, "".to_string(), false)?;
+                    /* int offset = 0;
+                     * for (i = 0; i < len(shapes); i++) {
+                     *     int max_i = shapes[i] - 1;
+                     *     int stride_i = strides[i];
+                     *     offset += max_i*stride_i;
+                     * }
+                     * int max_val = start + offset;
+                     * cmp max_val, data_size_ll_tmp
+                     */
```
deepakn94: This looks like dead code, but it's not? :)
Add a short comment above saying you're generating code for something similar to the following C code

weld/llvm.rs
```diff
@@ -806,53 +837,98 @@ impl LlvmGenerator {
...
+                    let el_ty = self.llvm_type(&Scalar(I64))?;
+                    /* both these have same type */
+                    let shapes_llvm_info = self.get_array_llvm_info(func, ctx, iter.shapes.as_ref().unwrap(), el_ty.clone(), false)?;
+                    let strides_llvm_info = self.get_array_llvm_info(func, ctx, iter.strides.as_ref().unwrap(), el_ty, false)?;
+                    let offset_ptr = ctx.var_ids.next();
+                    ctx.code.add(format!("{} = alloca i64", offset_ptr));
+                    ctx.code.add(format!("store i64 0, i64* {}", offset_ptr));
+                    let loop_name = "boundscheck_loop";
```
deepakn94: How are you currently executing these nested loops? This looks something like a while loop with some sort of `next()` method to get the next index to process on?

deepakn94:
At a high level, this looks great!

I left some minor comments / clarification questions. Would also be good to add some Python code that uses the `nditer`

deepakn94:
I'm not sure I understand your `size` question -- when are you trying to compute the size?

tests/integration_tests.rs
```diff
@@ -2260,6 +2260,133 @@ fn nested_for_loops() {
...
+    let code = "|x:vec[f64], shapes:vec[i64], strides:vec[i64]| result(for(nditer(x,0L,24L,1L,shapes,strides), appender, |b,i,e|
+                merge(b,log(e))))";
+
+    let conf = default_conf();
+    let mut x :[f64; 100] = [0.0; 100];
+    for i in 0..100 {
+        x[i] = i as f64;
+    }
+    let strides :[i64; 3] = [5, 2, 10];
+    let shapes :[i64; 3] = [2, 3, 4];
```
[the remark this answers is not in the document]
Parimarjan: So the original array (x) was of size 100 -- but here I was using shapes to construct a non-contiguous subarray of x of size 24 (so 2*3*4). The idea was that I used nditer to apply log(e) to the elements of this subarray, and then the I can use shapes/strides information to loop over just those elements of x and compare the results

weld/llvm.rs
Parimarjan: yup updated the comment above!

weld/llvm.rs
Parimarjan: I was just using a generic template in add_llvm_for_loop to generate the loop.start and loop.end code which handled incrementing the iteration variable etc. But I had a bug in this one - If there were multiple nditers, as in zip(nditer(e0,....), nditer(e1,....)) -- then there would be multiple for loops generated with the same name in llvm. I fixed this with the latest commit. Is this what you were referring to?

Parimarjan:
@deepakn94 thanks! I added a file with some of my tests in /examples/nditer -- which uses nditer from a python interface + a bug fix for the case with multiple nditers -- generated by zip(nditer(e0,...), nditer(e1,...)).
About the size, the size of the array being returned was being computed in transforms/infer_size as:
                                    let e = exprs::binop_expr(BinOpKind::Subtract,
                                                              *iters[0].end.as_ref().unwrap().clone(),
                                                              *start.clone())?;
                                    exprs::binop_expr(BinOpKind::Divide, e, *iters[0].stride.as_ref().unwrap().clone())?

which is essentially len = end-start / stride;
This is the only place in nditer where we need to use either end, or stride. So I was hoping to get rid of this and add code to computer len = shapes[0] x shapes[1] ... x shapes[n-1]. This should be fairly easy to do in weld code using: for(shapes, merger[i64, *], |b, i, e| merge(b, e));

But I wasn't sure how to generate this code using the ''exprs::'' syntax as used in the transforms/infer_size example above.

deepakn94:
Ah I see what you're saying w.r.t. the size computation. You should be able to generate that for loop using something pretty similar to this:

```
let new_loop = exprs::for_expr(
    iters.clone(),
    exprs::newbuilder_expr(Appender(ek.clone()), Some(length))?,
    func.as_ref().clone(),
    false)?;
```

You would need to replace `Appender` with `Merger(, BinOpKind::Multiply)`, and also wrap that above expression with a `result_expr`.

examples/python/nditer/nditer_test.py
```diff
@@ -0,0 +1,140 @@
...
+def test_view():
+    '''
+    Adding the iter code to a view.
+    In general, do we deal correctly with wn.array(view)??
+    '''
+    orig = np.random.rand(20,20)
+    a = orig[3:15:1,4:20:2]
+
+    # orig = np.random.rand(20,20,20)
+    # a = orig[3:15:3,:,:]
```
deepakn94: Remove dead code

deepakn94:
New tests look good -- can you also run these from Travis (you can see how we're currently calling stuff [here](https://github.com/weld-project/weld/blob/master/build_tools/travis/test_multi_version.sh#L35); should be as straight forward as running the new Python file)

parimarjan:
@deepakn94 @sppalkia this should be ready to be reviewed / merged....I didn't figure out the bug with the length calculation for nditer using the shapes array yet (the code I had for that is here: https://github.com/parimarjan/weld/blob/nditer_calc_length/weld/transforms/size_inference.rs#L82 ) . Currently, nditer's signature looks like: nditer(e0, start, end, stride, shape, strides)  -- here the end and stride variables ideally shouldn't be there (right now I just set end = length, and stride = 1 -- so they can be used to calculate length)

weld/type_inference.rs
```diff
@@ -464,6 +464,17 @@ fn infer_locally(expr: &mut PartialExpr, env: &mut TypeMap) -> WeldResult<bool>
...
                     }
                 }
+                /* For Nd-Iter */
+                if iter.shape.is_some() {
+                    for i in [&mut iter.shape, &mut iter.strides].iter_mut() {
+                        match **i {
+                            Some(ref mut e) => {
+                                changed |= try!(push_complete_type(&mut e.ty, Vector(Box::new(Scalar(I64))), "iter"))
+                            }
+                            None => return weld_err!("Impossible"),
```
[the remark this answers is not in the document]
parimarjan: fixed this

weld/llvm.rs
```diff
@@ -816,20 +826,60 @@ impl LlvmGenerator {
...
+        let first_iter = &par_for.data[0];
         let data_str = llvm_symbol(&first_data);
         let data_ty_str = self.llvm_type(func.params.get(&first_data).unwrap())?;
         let data_prefix = llvm_prefix(&data_ty_str);
-
         let num_iters_str = ctx.var_ids.next();
         let mut fringe_start_str = None;
-        if par_for.data[0].kind == IterKind::SimdIter || par_for.data[0].kind == IterKind::ScalarIter {
+        
+        if first_iter.kind == IterKind::NdIter {
```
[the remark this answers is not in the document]
parimarjan: yup switched to match statement here

weld/sir/mod.rs
```diff
@@ -771,6 +790,37 @@ pub fn ast_to_sir(expr: &TypedExpr, multithreaded: bool) -> WeldResult<SirProgra
...
+/// "shape" etc. Returns None, or the Symbol associated with the field. It also resets values for
+/// cur_func, and cur_block.
+fn get_iter_sym(opt : &Option<Box<Expr<Type>>>,
+            prog: &mut SirProgram,
+            cur_func: &mut FunctionId,
+            cur_block: &mut BasicBlockId,
+            tracker: &mut StatementTracker,
+            multithreaded: bool,
+            body_func: FunctionId) -> WeldResult<Option<Symbol>> {
+    if opt.is_some() {
```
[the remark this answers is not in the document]
parimarjan: fixed!

weld/transforms/size_inference.rs
```diff
@@ -47,11 +48,69 @@ fn func_has_simple_merge(expr: &TypedExpr) -> bool {
...
+                             * Note: to extract the let statements, let_expr needs the expression
+                             * in which the let statement is being defined. So first, we need to
+                             * define new_loop -- but to correctly define new loop, we require the
+                             * length -- thus we extract the let statements only at the end after
+                             * new loop has been defined.
+                             */
+                            let data_expr = ident_expr(data_sym.clone(), iters[0].data.ty.clone()).unwrap();
+                            (length_expr(data_expr.clone()).unwrap(), Some(data_expr))
+                        //} else if iters[0].kind == IterKind::RangeIter || iters[1].kind == IterKind::RangeIter {
+                        } else if iters.iter().any(|ref iter| iter.kind == IterKind::RangeIter) {
```
parimarjan: @sppalkia I wasn't sure what's the best way to deal with the RangeIter case - so just return None for now since it always works - but this should maybe be changed?

PR #266, opened 2017-11-06
Better dependency handling

deepakn94:


deepakn94:
@sppalkia @mateiz this should be ready to review.

I added a `grizzly.group()` primitive to tell the Weld compiler to co-optimize different operations. It takes in as input a list of `WeldObjects` and returns a new `WeldObject` that has dependencies on all the input `WeldObjects`.
Example usage:
```
       predicates = inp == 3
        output = gr.group(
            [predicates,
             inp.mask(predicates, 6)]
        ).evaluate()
```

I also tried to fix style formatting for the Python files I touched.

python/grizzly/grizzly/encoders.py
```diff
@@ -138,7 +138,7 @@ def __init__(self):
         lib_file = pkg_resources.resource_filename(__name__, lib)
         self.utils = ctypes.PyDLL(lib_file)
 
-    def decode(self, obj, restype):
+    def decode(self, obj, restype, raw_ptr=False):
         """Summary
 
         Args:
```
sppalkia: Can you add what `raw_ptr` does here

python/grizzly/grizzly/encoders.py
```diff
@@ -151,6 +151,12 @@ def decode(self, obj, restype):
         Raises:
             Exception: Description
         """
+        if raw_ptr:
+            data = obj
```
sppalkia: When is this used? When we pass back some partial result?

python/grizzly/grizzly/encoders.py
```diff
@@ -184,13 +190,18 @@ def decode(self, obj, restype):
             weld_to_numpy = self.utils.weld_to_numpy_double_arr_arr
         elif restype == WeldVec(WeldInt()):
             weld_to_numpy = self.utils.weld_to_numpy_int_arr
+        elif isinstance(restype, WeldStruct):
+            ret_vecs = []
+            for field_type in restype.field_types:
+                ret_vec = self.decode(data, field_type, raw_ptr=True)
```
sppalkia: ah it's for use here, got it :P

PR #269, opened 2017-11-09
Some LLVM Compile Time Improvements

sppalkia:
### Note: This change requires Rust >= 1.20

This PR makes some strides for improving compile times for LLVM. There are two main aspects:

1. Disable the module passes; it seems the LTO passes cover most of these, and there was no performance regression observed by only using them (we may revert this if we see anything that contradicts this).

2. Restructure the SIR to enable optimizations at the SIR level. This does the following:

- Change `Statement` to `StatementKind` and factor out the `output` `Symbol`. This allows recognizing unique "instructions" by just matching on the `StatementKind`.

We may eventually also want to make something like a `StatementValue`, which is an `enum` with `Symbol` and `LiteralKind`, which would allow inlining literals into LLVM expressions instead of alloca'ing them and load/storing them. All `StatementKind`s would then take `StatementValue` as input, and a `Statement` would output a `Symbol`. This is not yet implemented.

The main SIR level optimization here is creating the `StatementTracker`, which (conservatively) prevents emitting duplicate statements. It is conservative at the moment because it only works at the basic block level for simplicity, even though many expressions may be deduplicated across basic blocks within a function.

@jjthomas would be good for you to take a look at the SIR changes in particular.

### DIFF
--- a/weld/sir/optimizations/fold_constants.rs
+++ b/weld/sir/optimizations/fold_constants.rs
@@ -1,0 +1,193 @@
+//! Folds constants in SIR programs.
+//! 
+//! This transform walks each SIR function and attempts to evaluate expressions at runtime,
+//! deleting definitions and variables which evaluate to simple constants.
+//! 
+//! The transform in conservative in two ways:
+//! 
+//! 1. It will not remove symbols required by terminators, in case a code generator relies on the
+//! existence of these symbols.
+//! 
+//! 2. It will not remove symbols which are used as parameters to a function (though this can
+//!    probably be changed to support a simple inteprocedural constant simpliciation).
+
+use std::collections::BTreeMap;
+
+use error::*;
+use sir::*;
+
+extern crate fnv;
+
+/// Evaluates a binary operation over two literals, returning a new literal.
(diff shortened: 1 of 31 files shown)


jjthomas:
Looks good to me, I like the explicit adding of valid statements into the tracker.

sppalkia:
@jjthomas can you take a look again, this also implements the constant folding thing we talked about. Unfortunately it doesn't seem to help compile times too much yet but it's probably a good first step toward being able to disable some of the optimizations LLVM does by doing them on the SIR.

weld/sir/optimizations/fold_constants.rs
```diff
@@ -0,0 +1,193 @@
...
+                // If there's no output symbol (i.e., it's a `Merge`), retain it.
+                true
+            }
+        });
+    }
+
+    // TODO(shoumik): is there a better way of doing this...
+    let mut locals = BTreeMap::new();
+    for (k, v) in func.locals.iter() {
+        if used_symbols.contains(k) || global_params.contains(k) {
```
jjthomas: `global_params` may contain locals because you added terminator variables to it as well?

sppalkia: Yup!

PR #271, opened 2017-11-11
Add SIR Tracing as a Configuration Option

sppalkia:
This adds SIR tracing as a configurable option during compilation. This is based on @mateiz's earlier patch which did the same thing (but with the option to turn it off).

Also added some examples on how to set configuration options in the `add_repl` example.


mateiz:
Looks good to me.

PR #272, opened 2017-11-11
bugfix for nested maps with vecbuilder issue

jjthomas:
should resolve @rahulpalamuttam issues with birth analysis

### DIFF
--- a/weld_rt/cpp/vb.cpp
+++ b/weld_rt/cpp/vb.cpp
@@ -74,8 +74,9 @@ extern "C" void weld_rt_new_vb_piece(void *v, work_t *w, int32_t is_init_piece) 
     cur_piece->size = 0;
     cur_piece->capacity = vb->starting_cap;
   } else {
-    // if initial piece (created right after vb is created), w->cur_idx won't be set to the
-    // value we want (0)
+    // If this is the initial piece (i.e. the one created right after the vb is created),
+    // we want to use an offset of 0. Only for pieces created for new tasks should an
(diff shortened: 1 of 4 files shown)


weld_rt/cpp/vb.cpp
```diff
@@ -74,7 +74,11 @@ extern "C" void weld_rt_new_vb_piece(void *v, work_t *w) {
     cur_piece->size = 0;
     cur_piece->capacity = vb->starting_cap;
   } else {
-    cur_piece->data = (uint8_t *)vb->fixed_vector + vb->elem_size * w->cur_idx;
+    // If this is the initial piece (i.e. the one created right after the vb is created),
+    // we want to use an offset of 0. Only for pieces created for new tasks should an
+    // offset of w->cur_idx be used.
+    cur_piece->data = (uint8_t *)vb->fixed_vector +
+      (is_init_piece ? 0 : vb->elem_size * w->cur_idx);
```
rahulpalamuttam: More of a style thing here - but are we fine with using ternaries vs an explicit if else?

jjthomas: yeah i think in this case it's pretty straightforward and shortens the code, i agree in cases where you have nested ternaries and stuff it's pretty annoying

rahulpalamuttam:
Looks good.
We can also probably include the nested map test case. 
|v1: vec[vec[i8]]| map(v1, |v2:vec[i8]| map(v2, |e:i8| e))

sppalkia:
Thank you James!

PR #273, opened 2017-11-11
Add Nested Map Identity Test

sppalkia:
@rahulpalamuttam can you verify that birth analysis is now working? If not we can fix any other issues and add tests for it here as well.


rahulpalamuttam:
It's working!

sppalkia:
cool thanks for checking

PR #275, opened 2017-11-11
Short Circuiting Booleans

sppalkia:
This PR adds short circuit support for booleans. Note that this is a semantic change to the language.

As an example, before the patch, in the expression:

```
(false && expr)
```

`expr1` would be evaluated, but with this patch, it is not evaluated. Likewise, in `(true || expr)`, `expr` is no longer evaluated (but would have been earlier).

One exception is for `if` expressions with `@(predicate:true)` annotations. The _condition_ for the expression is then not short circuited, but any boolean expressions in the `on_true` and `on_false` subexpression are.

### DIFF
--- a/weld/transforms/short_circuit.rs
+++ b/weld/transforms/short_circuit.rs
@@ -98,6 +98,19 @@ fn complex_and_or() {
 }
 
 #[test]
+fn complex_and_or_2() {
+    let mut e = typed_expr("|x:i32| (x > 5 && x < 10) || (x == 15)");
+    short_circuit_booleans(&mut e);
+    let ref expect = typed_expr("|x:i32|
+                    if(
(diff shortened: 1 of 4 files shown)


weld/transforms/short_circuit.rs
```diff
@@ -0,0 +1,119 @@
...
+    short_circuit_booleans(&mut e);
+    let ref expect = typed_expr("|x:i32|
+                    if(
+                        if (x > 5, true, x < 10),
+                        if (x == 7, true, x == 2),
+                        false  
+                    )");
+    assert!(e.compare_ignoring_symbols(expect).unwrap());
+}
+
```
deepakn94: Could also add a `complete_or_and` test

deepakn94:
Looks good! Ship it!

PR #282, opened 2017-11-22
Better Pass Handling in Python

sppalkia:
This adds logging capabilities in Python, and also makes it easier to change which passes are run. Some required passes have been factored out to always run (e.g., `inline-apply`) -- users shouldn't specify these in the list of passes they want run.


deepakn94:
LGTM; thank you for this insightful patch!

PR #284, opened 2017-11-22
support for structs of appenders and structs of mergers

jjthomas:
also need to fix case with scalar mergers inside structs, but wanted to get this out right away


sppalkia:
Looks good, can you add a test with multicore structs of builders? I think the Black Scholes thing will need this fixed soon too for multicore.

jjthomas:
@sppalkia did this, also realized there's a bug exposed by complex_parallel_for_appender_loop when it actually runs in parallel, but I'll fix that in a subsequent PR

PR #286, opened 2017-11-24
New predication transform

deepakn94:
Added a new predicate to transform expressions from `if(cond, scalar1, scalar2)` to `select(cond, scalar1, scalar2)`.

Downstream vectorization doesn't work though.

### DIFF
--- a/weld/transforms/vectorizer.rs
+++ b/weld/transforms/vectorizer.rs
@@ -164,6 +164,29 @@ pub fn predicate_simple_expr(e: &mut Expr<Type>) {
             return Ok((None, true));
         }
 
+        // Check if any sub-expression has a builder; if so bail out in order to not break linearity.
+        let mut safe = true;
+        e.traverse(&mut |ref sub_expr| {
+            match sub_expr.kind {
+                Merge { .. } => {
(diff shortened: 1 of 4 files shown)


weld/transforms/vectorizer.rs
```diff
@@ -162,6 +162,26 @@ pub fn predicate(e: &mut Expr<Type>) {
...
+pub fn predicate_simple_expr(e: &mut Expr<Type>) {
+    e.transform_and_continue_res(&mut |ref mut e| {
+        if !(should_be_predicated(e)) {
+            return Ok((None, true));
+        }
+
+        // This pattern checks for if(cond, scalar1, scalar2).
+        if let If { ref cond, ref on_true, ref on_false } = e.kind {
+            if let Scalar(_) = on_true.ty {
+                if let Scalar(_) = on_false.ty {
```
deepakn94: @sppalkia Is this sufficient? If this contains a builder, then it must contain a result as well, and then it seems that this transformation is safe. Are there other scenarios I'm missing?

sppalkia: See the note I sent

PR #289, opened 2017-11-28
Loop size annotation

deepakn94:


python/grizzly/grizzly/numpy_weld_impl.py
```diff
@@ -105,16 +105,20 @@ def dot(matrix, vector, matrix_ty, vector_ty):
...
+        loopsize_annotation = "@(loopsize: %d)" % len(vector)
 
     weld_template = """
        map(
          %(matrix)s,
          |row: vec[%(matrix_ty)s]|
            result(
              for(
+               %(loopsize_annotation)s
                map(
```
sppalkia: Change this to a `result(for` for now? Otherwise the annotation won't hold.

PR #293, opened 2017-11-29
attempt at better hash finalize function

jjthomas:


jjthomas:
doesn't seem to provide improvement on some simple cases

jjthomas:
it actually works on some of the large Q1 tests

PR #298, opened 2017-12-21
fold_constants bugfix for symbols assigned multiple times

jjthomas:
@sppalkia we should add some SIR tests that cover this stuff at some point


sppalkia:
Looks okay, but why was this necessary? I thought the SIR uses SSA, so each symbol name is unique, right? Is there a bug elsewhere that’s causing them not to be? Might be better to fix that than to do this.

jjthomas:
'If' nodes cause symbols to be assigned multiple times -- the result must be assigned in both the true and false block. If we want to make this SSA we'll have to introduce LLVM-like phi nodes, which is more complicated. I think we should just leave it as is. If pure SSA really helps us for some transform we can think about a design for that.

sppalkia:
Ah got it! Looks good then, you can merge it. We can think about the Phi
thing later but not necessary for now.

PR #300, opened 2018-01-15
Add StringLiteral to LiteralKind

pratiksha:
Plus, modify separator for Symbols to facilitate parsing from pretty printed functions.

### DIFF
--- a/weld/exprs.rs
+++ b/weld/exprs.rs
@@ -31,7 +31,7 @@ pub fn literal_expr(kind: LiteralKind) -> WeldResult<Expr<Type>> {
                  U64Literal(_) => Scalar(ScalarKind::U64),
                  F32Literal(_) => Scalar(ScalarKind::F32),
                  F64Literal(_) => Scalar(ScalarKind::F64),
-                 StringLiteral(_) => Vector(Box::new(Scalar(ScalarKind::U8))),
+                 StringLiteral(_) => Vector(Box::new(Scalar(ScalarKind::I8))),
              })
 }
 
--- a/weld/exprs.rs
+++ b/weld/exprs.rs
@@ -18,20 +18,21 @@ pub fn new_expr(kind: ExprKind<Type>, ty: Type) -> WeldResult<Expr<Type>> {
 }
 
 pub fn literal_expr(kind: LiteralKind) -> WeldResult<Expr<Type>> {
-    new_expr(Literal(kind),
-             Scalar(match kind {
-                        BoolLiteral(_) => ScalarKind::Bool,
-                        I8Literal(_) => ScalarKind::I8,
-                        I16Literal(_) => ScalarKind::I16,
-                        I32Literal(_) => ScalarKind::I32,
-                        I64Literal(_) => ScalarKind::I64,
-                        U8Literal(_) => ScalarKind::U8,
-                        U16Literal(_) => ScalarKind::U16,
-                        U32Literal(_) => ScalarKind::U32,
-                        U64Literal(_) => ScalarKind::U64,
-                        F32Literal(_) => ScalarKind::F32,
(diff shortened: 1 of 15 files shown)


weld/llvm.rs
```diff
@@ -1908,6 +1937,24 @@ impl LlvmGenerator {
         Ok(())
     }
 
+    fn string_literal(&mut self, string: &String, vec_ty: &String, ctx: &mut FunctionContext) -> WeldResult<String> {
```
sppalkia: Same as above

weld/llvm.rs
```diff
@@ -2629,8 +2679,16 @@ impl LlvmGenerator {
                 if let Simd(_) = *output_ty {
                     self.gen_simd_literal(&output_ll_sym, value, output_ty, ctx)?;
                 } else {
-                    let ref value = llvm_literal(*value);
-                    self.gen_store_var(value, &output_ll_sym, &output_ll_ty, ctx);
+                    match value {
```
sppalkia: do `match *value` and get rid of the `&` on the next line

weld/llvm.rs
```diff
@@ -3386,8 +3444,8 @@ fn llvm_lt(k: ScalarKind) -> &'static str {
 }
 
 /// Returns an LLVM formatted String for a literal.
-fn llvm_literal(k: LiteralKind) -> String {
-    match k {
+fn llvm_literal(k: LiteralKind) -> WeldResult<String> {
```
sppalkia: Might make more sense to move this into `LlvmGenerator` as a function now, and then call `gen_string_definition` etc. when matching on `StringLiteral`?

weld/exprs.rs
```diff
@@ -18,20 +18,21 @@ pub fn new_expr(kind: ExprKind<Type>, ty: Type) -> WeldResult<Expr<Type>> {
...
+                 I16Literal(_) => Scalar(ScalarKind::I16),
+                 I32Literal(_) => Scalar(ScalarKind::I32),
+                 I64Literal(_) => Scalar(ScalarKind::I64),
+                 U8Literal(_) => Scalar(ScalarKind::U8),
+                 U16Literal(_) => Scalar(ScalarKind::U16),
+                 U32Literal(_) => Scalar(ScalarKind::U32),
+                 U64Literal(_) => Scalar(ScalarKind::U64),
+                 F32Literal(_) => Scalar(ScalarKind::F32),
+                 F64Literal(_) => Scalar(ScalarKind::F64),
+                 StringLiteral(_) => Vector(Box::new(Scalar(ScalarKind::U8))),
```
sppalkia: In type inference below this pushed type is `I8` (instead of `U8`), and in the test above it's cast into a `u8`. Make sure it's consistent (I'd make it `i8` everywhere, so change this and the test above to cast to `i8`).

weld/type_inference.rs
```diff
@@ -131,6 +131,9 @@ fn infer_locally(expr: &mut PartialExpr, env: &mut TypeMap) -> WeldResult<bool>
 
         Literal(BoolLiteral(_)) => push_complete_type(&mut expr.ty, Scalar(Bool), "BoolLiteral"),
 
+        Literal(StringLiteral(_)) => push_complete_type(&mut expr.ty, Vector(Box::new(Scalar(I8))),
```
sppalkia: See comment above - make sure everything else is `i8 instead of `u8`.

weld/llvm.rs
pratiksha: I think I'm actually going to keep these separate, because e.g. `string_literal` requires passing in function state while `llvm_literal` doesn't.

weld/type_inference.rs
pratiksha: done

pratiksha:
@sppalkia

PR #304, opened 2018-02-06
New Workload, add Transforms

sppalkia:
[TODO - fix description]


weld/transforms/inliner.rs
```diff
@@ -122,6 +122,154 @@ pub fn inline_let(expr: &mut Expr<Type>) {
...
 }
 
+/// Changes negations of literal values to be literal negated values.
+pub fn inline_negate(expr: &mut TypedExpr) {
+    use ast::LiteralKind::*;
+    use exprs::literal_expr;
+    expr.transform(&mut |ref mut expr| {
+        if let Negate(ref child_expr) = expr.kind {
+            if let Literal(ref literal_kind) = child_expr.kind {
+                let res = match  *literal_kind {
```
hvanhovell: Should you deal with overflows here? For example `-(-2147483648)`?

PR #309, opened 2018-02-25
RangeIter implementation

sppalkia:
Implements a basic iterator over a range of values without materializing a vector.

The syntax for this follows the other iterators:

`for(rangeiter(0L, 100L, 1L), ...)` iterates over the interval `[0, 100)` and accesses every element. The arguments are the start index, the end index, and the stride; iteration continues until the index is greater than or equal to the specified end index.

The range iterator follows all the standard rule of the for loop: namely, the number of iterations must be equal across all zipped iterators.

## Limitations

* Currently doesn't support transforms that inspect iterators (namely, loop fusion and vectorization).

### DIFF
--- a/examples/cpp/vector_benchmark/vector_benchmark.cpp
+++ b/examples/cpp/vector_benchmark/vector_benchmark.cpp
@@ -17,8 +17,8 @@ struct args {
     int32_t a;
 };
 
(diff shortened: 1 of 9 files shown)


jjthomas:
Looks good

PR #325, opened 2018-03-19
Fix grizzly clang path

radujica:


sppalkia:
Thanks for the change!

PR #359, opened 2018-05-23
Add apply_once option to terminate transform early

pratiksha:
Not used in any of the transforms currently in master, but this keeps coming up in experimental features (cost model, RPC).


sppalkia:
@pratiksha I'm closing this since it's not used anywhere right now.

PR #372, opened 2018-07-02
[WIP] LLVM Single-threaded Backend

sppalkia:
This is a single threaded backend using the LLVM API. It is still a work in progress. Updated comments coming soon.


weld/codegen/llvm2/mod.rs
```diff
@@ -0,0 +1,1107 @@
...
+mod serde;
+mod vector;
+
+use self::builder::merger;
+use self::builder::appender;
+
+pub fn compile(program: &SirProgram,
+               conf: &ParsedConf,
+               stats: &mut CompilationStats,
+               dump_prefix: &str) -> WeldResult<Box<dyn Runnable>> {
```
mateiz: As a general note, it would be good to have a brief comment on every function and struct. Helps a lot with following the code later!

PR #386, opened 2018-08-25
Make pyweld Python 3 friendly

radujica:
- use `__future__ print_function` & update prints
- optimize imports
- use `encode` with `c_char_p`
- move to separate static method the weld input name generation
- use `sorted`

Tested with Python 2.7 & 3.5:

    python setup.py bdist_wheel
    python3 setup.py bdist_wheel

    import numpy as np

    from weld.encoders import NumpyArrayDecoder, NumpyArrayEncoder
    from weld.types import WeldLong
    from weld.weldobject import WeldObject

    weld_obj = WeldObject(NumpyArrayEncoder(), NumpyArrayDecoder())
    obj_id = weld_obj.update(np.array([1, 2, 3]))
    weld_obj.weld_code = "len(%s)" % obj_id

    print(weld_obj.evaluate(WeldLong()))

### DIFF
--- a/python/pyweld/setup.py
+++ b/python/pyweld/setup.py
@@ -2,10 +2,9 @@ import os
 import platform
 import shutil
 import subprocess
-import sys
 
-from setuptools import setup, Distribution
 import setuptools.command.build_ext as _build_ext
+from setuptools import setup, Distribution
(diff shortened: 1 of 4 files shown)


sppalkia:
Thanks Radu!

PR #387, opened 2018-08-26
Serde fixes

sppalkia:
Restructures serialization code so that the logic for each type is separated into its own function. This makes control flow within a serialization routine much simpler. It also fixes an issue where nested types with control flow in their routines (e.g., `deserialize[i32, vec[vec[u8]]](...)` would fail.

## Other Issues Resolved in this Patch

* Building the `cdylib` was accidentally disabled, so it was not possible to run the C++ examples. This is fixed.
* An issue where dumping code fails if two programs are compiled within the same second has been fixed, by adding a unique suffix to the timestamp if a file with the given timestamp already exists.

## Tasks for this Patch

* Add a test for deserializing a type such as `dict[i32,vec[vec[u8]]]` (where the value is itself a nested collection type).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,7 @@ uuid = { version = "0.6.5", features = ["v4"] }
 [lib]
 path = "weld/lib.rs"
 bench = false
+crate-type = ["cdylib", "rlib"]
 
 [[bin]]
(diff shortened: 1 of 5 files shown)


abehm:
lgtm

PR #403, opened 2018-10-10
Common Subexpression Elimination

sppalkia:
Implements the common subexpression elimination transformation.

### DIFF
--- a/weld/optimizer/transforms/cse.rs
+++ b/weld/optimizer/transforms/cse.rs
@@ -232,7 +232,7 @@ impl Site {
 
     /// Returns whether this site contains `other`.
     ///
-    /// A site contains `a` contains another site `b` if `a` is a substring of `b`.
+    /// A site contains `a` contains another site `b` if `a` is a prefix of `b`.
     ///
     /// Returns `true` if the two sites are equal.
     fn contains(&self, other: &Site) -> bool {
@@ -334,29 +334,11 @@ impl Cse {
 
         cse.counter = 0;
         let ref mut sites =  cse.build_site_map(expr, bindings);
-
-        // Debug.
-        {
-            let mut debug = String::new();
-            for (k, v) in bindings.iter() {
-                debug.push_str(&format!("{} -> {}\n", k, v.pretty_print()));
-            }
-            trace!("bindings: {}", debug);
-            trace!("Expression after assigning cse symbols: {}", expr.pretty_print());
-             let mut debug = String::new();
-            for (k, v) in sites.iter() {
-                debug.push_str(&format!("{} -> {:?}\n", k, v));
-            }
-            trace!("sites: {}", debug);
-        }
-
         let ref mut generated = HashSet::new();
         let ref mut stack = vec![];
 
         cse.counter = 0;
         cse.generate_bindings(expr, bindings, generated, &mut Site::new(), stack, sites);
-
-        trace!("After CSE: {}", expr.pretty_print());
     }
 
     /// Removes common subexpressions and remove bindings.
@@ -373,8 +355,6 @@ impl Cse {
                                     aliases: &mut HashMap<Symbol, Vec<Symbol>>) {
 
         use self::UseCse;
-
-        trace!("Remove Common Subexpressions called on {}", expr.pretty_print());
 
         expr.transform_up(&mut |ref mut e| {
 
@@ -607,11 +587,6 @@ impl Cse {
                                stack: &mut Vec<Vec<(Symbol, Expr)>>,
                                sites: &mut SiteMap) {
 
-        trace!("Processing scoped expression {} (old site={:?}, counter={}",
-               expr.pretty_print(),
-               current_site,
-               self.counter);
-
         current_site.push(self.counter);
         self.counter += 1; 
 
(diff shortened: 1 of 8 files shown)


weld/optimizer/transforms/cse.rs
```diff
@@ -0,0 +1,561 @@
...
+
+    /// Generates bindings in the expression.
+    ///
+    /// This injects the definition of each symbol into the correct position in the final
+    /// expression. Specifically, a symbol is defined in the highest scope that symbol is used in.
+    ///
+    /// The function updates `expr` with the symbol to value mappings in `bindings`. Bindings that
+    /// are already inserted are tracked in `generated`. The `stack` tracks scopes to generate
+    /// the bindings in, and `scopes` maps each symbol to its highest scope.
+    fn generate_bindings(&mut self,
```
deepakn94: This function is a little hard to understand :(

weld/optimizer/transforms/cse.rs
```diff
@@ -0,0 +1,561 @@
...
+//! generation_.
+//!
+//! # 1. Subexpression Elimination
+//!
+//! In subexpression elimination, first, we create a hash table that maps each expression to a
+//! symbol name. The hash table is created by traversing the expression tree bottom-up (from leaf
+//! to root), so sub-expressions are added to the table before parent expressions. When processing
+//! an expression, we check if its exists in the hash table, and if it does, we replace it with the
+//! corresponding symbol.
+//!
```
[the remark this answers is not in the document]
sppalkia: Yeah, it's breath first post-order traversal (see the `transform_up` method in `ast.rs`.

For the second question, the expression tree is supposed to be uniquified, so symbol names are always unique (incidentally I forgot to call `uniquify`, so good catch :))

weld/optimizer/transforms/cse.rs
```diff
@@ -0,0 +1,561 @@
...
+//! Calling the inliner will result in a more sane-looking expression tree:
+//!
+//! ```weld
+//! let cse0 = lookup(w, 1);
+//! if (x > 0,
+//!     let cse1 = lookup(v, 0); cse1 + cse0,
+//!     lookup(z, 0)
+//! ) + cse0
+//! ```
```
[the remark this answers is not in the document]
sppalkia: ah good point, it's not pedantic, it would be wrong if it reordered expressions :)

weld/optimizer/transforms/cse.rs
```diff
@@ -0,0 +1,561 @@
...
+
+            if !e.use_cse() {
+                return None;
+            }
+
+            let taken;
+            match e.kind {
+                Let { ref name, ref mut value, ref mut body } => {
+                    // Let expressions represent "already CSE'd" values, so we can track them.
+                    let taken_value = value.take();
```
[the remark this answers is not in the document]
sppalkia: See `trait Takeable` in `ast.rs`: it basically replaces an expression in a tree with a place holder and takes ownership of the expression that was there before, so you can edit it without borrow checker errors.

weld/optimizer/transforms/cse.rs
```diff
@@ -0,0 +1,573 @@
...
+        // we can CSE this by choosing the "outermost" alias and using it everywhere.
+        for (sym, alias_list) in aliases.drain() {
+            for alias in alias_list {
+                let expr = bindings.get(&sym).cloned().unwrap();
+                bindings.insert(alias, expr);
+            }
+        }
+
+        let ref scopes =  cse.build_scope_map(expr, bindings);
+        cse.generate_bindings(expr, bindings, &mut HashSet::new(), &mut vec![], scopes);
```
deepakn94: Comment here on what `HashSet::new()` and `vec![]` correspond to (`generated` and `stack`)

weld/optimizer/transforms/cse.rs
```diff
@@ -0,0 +1,573 @@
...
+        trace!("Processing scoped expression {}", expr.pretty_print());
+
+        stack.push(vec![]);
+
+        self.generate_bindings(expr, bindings, generated, stack, scopes);
+
+        // Generate the Let expressions for the bindings in this scope.
+        let binding_list = stack.pop().unwrap();
+        let mut prev = expr.take();
+        for (sym, expr) in binding_list.into_iter().rev() {
```
deepakn94: Comment on why you're doing this in reverse order. I had to think a while why you did this (because you put all "previous" let expressions in the body of the `let_expr` so they're actually executed after the current let expression?)

weld/optimizer/transforms/cse.rs
```diff
@@ -0,0 +1,573 @@
...
+    }
+
+    /// Generates bindings in the expression.
+    ///
+    /// This injects the definition of each symbol into the correct position in the final
+    /// expression. Specifically, a symbol is defined in the highest scope that symbol is used in.
+    ///
+    /// The function updates `expr` with the symbol to value mappings in `bindings`. Bindings that
+    /// are already inserted are tracked in `generated`. The `stack` tracks scopes to generate
+    /// the bindings in, and `scopes` maps each symbol to its highest scope.
```
deepakn94: I would add an explicit statement about what the scopes in the stack mean (list of (symbol, expression) mappings) as opposed to what's in the `scopes` map (which is a symbol to `scope_id` mapping)

PR #415, opened 2018-11-10
Algebraic optimizations and not operator

sppalkia:
@Bathtor these are your changes! Would you like to adopt this branch and submit a PR so the commits will go under your name? You can create a patch file from this PR (just add `.diff` to the end of the URL), apply it to the `llvm-st` branch of `weld-project/weld` with `git apply`, and submit a new pull request to replace this one if you'd like.

Since the this supersedes the one you originally posted, I'll go ahead and close that.


sppalkia:
(See #417)

PR #355, opened 2018-05-15
Added operator ! and an algebraic optimisations pass

Bathtor:
I tried to sort out the mess rustfmt made into separate commits as best as possible, but 8584dbc858fa244367babbaf5fc11d8a20af3ce4 still has a bunch of formatting changes, that are difficult to completely factor out.

PR #417, opened 2018-11-10
Algebraic optimizations and not operator cleaned up

Bathtor:
Cleaned up version of #355 based on #415.

Thanks @sppalkia for cleaning up the mess of the original :)


sppalkia:
Great thanks! Glad to finally merge this :)

PR #421, opened 2018-11-22
Sort for vectors and structs, plus custom comparators

pratiksha:
:turkey: 

- Replace `keyfunc` with custom comparator
- Add `compare` macro to standard macros
- Add default comparator BinOps (<, <=, >, >=) for vectors and structs
- Implement sort call


weld/codegen/llvm2/cmp.rs
```diff
@@ -0,0 +1,420 @@
...
+    /// The comparator should return -1 for left < right, 1 for left > right, and 0 for left == right.
+    /// These functions are not currently cached.
+    unsafe fn gen_custom_cmp(&mut self,
+                             elem_ty: &LLVMTypeRef,
+                             cf_id: &FunctionId,
+                             cmpfunc: &LLVMValueRef) -> WeldResult<LLVMValueRef>;
+}
+
+impl GenCmp for LlvmGenerator {
+    /// Generates a default comparison function for a type.
```
sppalkia: Newline after this

weld/sir/mod.rs
```diff
@@ -1065,30 +1065,30 @@ fn gen_expr(expr: &Expr,
...
                        ref params,
                        ref body,
-            } = keyfunc.kind {
-                let keyfunc_id = prog.add_func();
-                let keyblock = prog.funcs[keyfunc_id].add_block();
-                let (keyfunc_id, keyblock, key_sym) = gen_expr(body, prog, keyfunc_id, keyblock, tracker)?;
+            } = cmpfunc.kind {
+                let cmpfunc_id = prog.add_func();
+                let cmpblock = prog.funcs[cmpfunc_id].add_block();
+                let (cmpfunc_id, cmpblock, cmp_sym) = gen_expr(body, prog, cmpfunc_id, cmpblock, tracker, multithreaded)?;
```
sppalkia: Some small compile errors here :)

weld/codegen/llvm2/mod.rs
```diff
@@ -1245,8 +1271,60 @@ impl LlvmGenerator {
...
+                    let mut args = vec![elems_ptr, size, ty_size, comparator, run];
+                    let mut arg_tys = vec![LLVMTypeOf(elems_ptr),
+                                           LLVMTypeOf(size),
+                                           LLVMTypeOf(ty_size),
+                                           LLVMTypeOf(comparator),
+                                           LLVMTypeOf(run)];
+                    
+                    // Generate the call to qsort
+                    let void_type = self.void_type();
+                    self.intrinsics.add("qsort_r", void_type, &mut arg_tys); // In-place sort.
```
sppalkia: Hmm, this should technically return a new vector, which means we should clone the first one before sorting...unfortunately we don't have code to recursively clone stuff right now, which makes this quite annoying. Maybe add a note about it as a TODO and we can put it in as a patch later on.

tests/sort_tests.rs
```diff
@@ -136,7 +132,7 @@ fn complex_sort() {
                         |b,i,e| merge(b, e)
                       )
                     ),
-                    |x:{i32, i32}| x.$0
+                    |x:{i32, i32}, y:{i32, i32}| compare(x.$0, y.$0)
```
sppalkia: Maybe swap the zip order so the sort actually does something? I think right now its already sorted by the first struct element, so nothing will change.

sppalkia: Probably worth adding a few more tests here -- e.g., vectors of floats to make sure memcmp isn't called at the wrong time, structs with mixed types, structs with structs or vectors, etc. I can write some of these as well

weld/codegen/llvm2/mod.rs
```diff
@@ -1245,8 +1271,60 @@ impl LlvmGenerator {
...
                 }
             }
-            Sort { .. } => {
-                unimplemented!() // Sort
+            Sort { ref child, ref cmpfunc } => { // cmpfunc is an SirFunction
+                let output_pointer = context.get_value(output)?;
+                let output_type = context.sir_function.symbol_type(
+                    statement.output.as_ref().unwrap())?;
+                
+                let ref cmpfunc_func = context.sir_program.funcs[*cmpfunc];
```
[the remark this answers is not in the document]
pratiksha: I think this particular line isn't needed, but not sure what you mean about skipping the forward declare?

tests/sort_tests.rs
pratiksha: Added a few tests.

weld/codegen/llvm2/mod.rs
```diff
@@ -1245,8 +1259,60 @@ impl LlvmGenerator {
...
+                    use ast::Type::Scalar;
+                    use ast::ScalarKind;
+                    use self::vector::VectorExt;
+                    let zero = self.zero(self.i64_type());
+                    let child_elems = self.gen_at(context.builder, child_type, child_value, zero)?;
+                    let elems_ptr = LLVMBuildBitCast(context.builder, child_elems,
+                                                     self.void_pointer_type(),
+                                                     c_str!(""));
+                    let size = self.gen_size(context.builder, child_type, child_value)?;
+                    let mut elem_ll_ty = self.llvm_type(&*elem_ty)?;
```
sppalkia: Doesn't need to be`mut`. Also can remove `&*`

weld/codegen/llvm2/mod.rs
```diff
@@ -1245,8 +1259,60 @@ impl LlvmGenerator {
...
+                    let ty_size = self.size_of(elem_ll_ty);
+                    
+                    use self::cmp::GenCmp;
+                    let cmpfunc_ll_fn = self.functions[cmpfunc];
+
+                    let run = context.get_run();
+
+                    // Generate the comparator from the provided custom code.
+                    // The custom comparator can only make use of BinOps that are supported by Weld
```
sppalkia: Maybe delete this comment? Since you can also have `If`, `GetField`, etc etc

weld/codegen/llvm2/numeric.rs
```diff
@@ -262,13 +264,41 @@ impl NumericExpressionGen for LlvmGenerator {
...
+                            let gt = LLVMBuildICmp(ctx.builder, LLVMIntSGT, cmp,
+                                                   self.i32(0), c_str!(""));
+                            
+                            if op == BinOpKind::GreaterThan {
+                                gt
+                            } else {
+                                LLVMBuildNot(ctx.builder, gt, c_str!(""))
+                            }
+                        }, 
                         _ => unimplemented!(), // Vector comparison
```
sppalkia: Remove comment and change to `unreachable!()` since all comparison cases are covered here.

weld/codegen/llvm2/cmp.rs
```diff
@@ -0,0 +1,351 @@
...
+    ///
+    /// For scalars, the comparison function returns -1 if left < right, 1 if left > right,
```
sppalkia: I think if you want to return the result of memcmp (which returns the difference between the differing bytes), we should change this to say "returns less than zero of left < right, greater than zero if left > right, and 0 if they are equal."

weld/codegen/llvm2/cmp.rs
```diff
@@ -0,0 +1,351 @@
...
+                let name = "memcmp";
+                let ret_ty = self.i32_type();
+                let ref mut arg_tys = [self.void_pointer_type(), self.void_pointer_type(), self.i64_type()];
+                if self.intrinsics.add(name, ret_ty, arg_tys) {
+                    let memcmp = self.intrinsics.get(name).unwrap();
+                    LLVMExtAddAttrsOnParameter(self.context, memcmp, &[ReadOnly, NoCapture], 0);
+                    LLVMExtAddAttrsOnParameter(self.context, memcmp, &[ReadOnly, NoCapture], 1);
+                }
+                let ref mut args = [left_data, right_data, bytes];
+                let memcmp_result = self.intrinsics.call(builder, name, args)?;
```
sppalkia: This is a little weird because it'll return things like `'a' == 'abc'`, which we don't want. I think if the bytes are equal, you need to compare the sizes to decide the final result. Adding this code (+/- typos) should fix it:

```rust
let func = self.gen_cmp_fn(&Scalar(I64))?;
let mut args = [left_size, right_size];
let size_eq = LLVMBuildCall(builder, func, args.as_mut_ptr(), args.len() as u32, c_str!(""));

let bytes_equal = LLVMBuildICmp(builder, LLVMIntEQ, memcmp_result, self.i32(0), c_str!(""));
let result = LLVMBuildSelect(builder, bytes_equal, size_eq, memcmp_result, c_str!(""));
LLVMBuildRet(builder, result);
```

You should add a test that compares vectors like [1] and [1, 2, 3, 4], where one is a substring of another.

weld/codegen/llvm2/cmp.rs
```diff
@@ -0,0 +1,351 @@
...
+
+                LLVMBuildRet(builder, memcmp_result);
```
sppalkia: (delete line after adding code above)

weld/codegen/llvm2/cmp.rs
```diff
@@ -0,0 +1,351 @@
...
+
+                let func = self.gen_cmp_fn(elem)?;
+                let left_value = self.gen_at(builder, ty, left_vector, phi_i)?;
+                let right_value = self.gen_at(builder, ty, right_vector, phi_i)?;
+                let mut args = [left_value, right_value];
+                let cmp_result = LLVMBuildCall(builder, func, args.as_mut_ptr(), args.len() as u32, c_str!(""));
+                let neq = LLVMBuildICmp(builder, LLVMIntNE, cmp_result, self.i32(0), c_str!(""));
+
+                let updated_i = LLVMBuildNSWAdd(builder, phi_i, self.i64(1), c_str!(""));
+                let check1 = LLVMBuildICmp(builder, LLVMIntEQ, updated_i, left_size, c_str!(""));
```
sppalkia: This should compare against `min_size`, not `left_size`

weld/codegen/llvm2/cmp.rs
```diff
@@ -0,0 +1,351 @@
...
+
+                let mut blocks = [entry_block, loop_block];
+                let mut values = [self.i64(0), updated_i];
+                LLVMAddIncoming(phi_i, values.as_mut_ptr(), blocks.as_mut_ptr(), values.len() as u32);
+
+                // Finish block - set result and compute number of consumed bits.
+                LLVMPositionBuilderAtEnd(builder, done_block);
+                let result = LLVMBuildPhi(builder, self.i32_type(), c_str!(""));
+                let mut blocks = [entry_block, loop_block];
+                let mut values = [self.i32(1), cmp_result];
```
sppalkia: I'd change `self.i32(1)` to `self.i32(0)`, since in the memcmp version, if there are no bytes to compare, `memcmp` will return 0 (and then the size check happens).

weld/codegen/llvm2/cmp.rs
```diff
@@ -0,0 +1,351 @@
...
+                LLVMAddIncoming(result, values.as_mut_ptr(), blocks.as_mut_ptr(), values.len() as u32);
+
```
sppalkia: Add the size check:

```rust

let func = self.gen_cmp_fn(&Scalar(I64))?;
let mut args = [left_size, right_size];
let size_eq = LLVMBuildCall(builder, func, args.as_mut_ptr(), args.len() as u32, c_str!(""));

let bytes_equal = LLVMBuildICmp(builder, LLVMIntEQ, result, self.i32(0), c_str!(""));
let result = LLVMBuildSelect(builder, bytes_equal, size_eq, result, c_str!(""));
```

weld/codegen/llvm2/cmp.rs
```diff
@@ -0,0 +1,351 @@
...
+        // Free the allocated string.
+        LLVMDisposeMessage(c_prefix);
+
+        let (function, builder, _) = self.define_function_with_visibility(ret_ty,
+                                                                          &mut arg_tys,
+                                                                          LLVMLinkage::LLVMExternalLinkage,
+                                                                          name);
+
+        LLVMExtAddAttrsOnFunction(self.context, function, &[InlineHint]);
```
sppalkia: Delete this since it shouldn't be inlined (it's only purpose is to be passed to a libc function)

weld/codegen/llvm2/cmp.rs
```diff
@@ -0,0 +1,351 @@
...
+    /// function returns -1 if the first element is smaller, 0 if the elements are equal, and 1 if
+    /// the first element is larger.
+    unsafe fn gen_cmp_fn(&mut self, ty: &Type) -> WeldResult<LLVMValueRef>; 
+
+    /// Generates an opaque comparator for a type.
+    ///
+    /// Opaque comparators are linked externally and operate over opaque (i.e., `void*`)
+    /// arguments. These are used, e.g., as an argument to qsort. This method returns the
+    /// generated function.
+    unsafe fn gen_opaque_cmp_fn(&mut self, ty: &Type) -> WeldResult<LLVMValueRef>;
```
sppalkia: Delete this declaration and its implementation: I actually don't think this is necessary since you have the `gen_custom_cmp` thing, which does all the visibility stuff (it's also not used anywhere) since gen_custom_cmp will internally generate the `cmp` code and wrap it to be externally visible already.

weld/codegen/llvm2/cmp.rs
```diff
@@ -0,0 +1,351 @@
...
+        let result = LLVMBuildZExt(builder, result, self.i32_type(), c_str!(""));
+        LLVMBuildRet(builder, result);
+        LLVMDisposeBuilder(builder);
+
+        self.opaque_cmp_fns.insert(ty.clone(), function);
+        Ok(function)
+    }
+
+    unsafe fn gen_custom_cmp(&mut self,
+                             elem_ty: &LLVMTypeRef,
```
sppalkia: nit: Since all the LLVM `Ref` types implement `Copy`, don't pass them around as references

weld/codegen/llvm2/cmp.rs
```diff
@@ -0,0 +1,351 @@
...
+                             cf_id: &FunctionId,
```
sppalkia: Same for FunctionId, it's just typedef's as an integer so no need for `&`

tests/binop_tests.rs
```diff
@@ -0,0 +1,76 @@
...
+    
+    let ret_value = compile_and_run(code, conf, input_data);
+    let data = ret_value.data() as *const bool;
+    let result = unsafe { *data };
+
+    assert!(result == false);
+}
+
+#[test]
+fn vector_cmp() {
```
sppalkia: Also add a test that compares vectors that are substrings of one another, e.g., `[1] == [1, 2, 3]`. Currently, that comparison will return `true`: I think you can add the code I put in the comments below to fix it.

weld/codegen/llvm2/cmp.rs
pratiksha: Sure, wasn't sure if we wanted to leave it in for potential future external calls

weld/codegen/llvm2/mod.rs
pratiksha: The point was you can't use BinOps on dicts or something where there isn't a default comparator, but I see how this comment is confusing

weld/codegen/llvm2/cmp.rs
```diff
@@ -0,0 +1,323 @@
...
+                    
+                    LLVMPositionBuilderAtEnd(builder, next_block);
+                }
+                
+                // All blocks equal, return 0.
+                LLVMBuildRet(builder, result); 
+            }
+            // Vectors comprised of unsigned chars or booleans can be compared for using `memcmp`.
+            Vector(ref elem) if elem.supports_memcmp() => {
+                let left_vector = self.load(builder, left)?;
```
sppalkia: Sorry to nit, but can you actually change this to:

```rust
let left_data_ptr = LLVMBuildStructGEP(builder, left, vector::POINTER_INDEX, c_str!(""));
let left_data = self.load(builder, left_data_ptr)?;
let right_data_ptr = LLVMBuildStructGEP(builder, right, vector::POINTER_INDEX, c_str!(""));
let left_data = self.load(builder, right_data_ptr)?;

let left_size_ptr = LLVMBuildStructGEP(builder, left, vector::SIZE_INDEX, c_str!(""));
let right_size_ptr = LLVMBuildStructGEP(builder, right, vector::SIZE_INDEX, c_str!(""));
```

The reason for this is that adding alloca's in the middle of a function (as opposed to the very top) confuses the LLVM optimizer, and it sometimes can't figure out that it needs to move things to registers.

weld/codegen/llvm2/cmp.rs
```diff
@@ -0,0 +1,323 @@
...
+                    let memcmp = self.intrinsics.get(name).unwrap();
+                    LLVMExtAddAttrsOnParameter(self.context, memcmp, &[ReadOnly, NoCapture], 0);
+                    LLVMExtAddAttrsOnParameter(self.context, memcmp, &[ReadOnly, NoCapture], 1);
+                }
+                let ref mut args = [left_data, right_data, bytes];
+                let memcmp_result = self.intrinsics.call(builder, name, args)?;
+
+                // If all compared bytes were equal but sizes were not equal, the smaller vector is the lesser element.
+                let func = self.gen_cmp_fn(&Scalar(I64))?;
+                let left_size_ptr = LLVMBuildAlloca(builder, self.i64_type(), c_str!(""));
```
sppalkia: Delete these and use the suggested lines above to avoid the alloca

weld/codegen/llvm2/cmp.rs
```diff
@@ -0,0 +1,323 @@
...
+                // Finish block - set result and compute number of consumed bits.
+                LLVMPositionBuilderAtEnd(builder, done_block);
+                let result = LLVMBuildPhi(builder, self.i32_type(), c_str!(""));
+                let mut blocks = [entry_block, loop_block];
+                let mut values = [self.i32(0), cmp_result];
+                LLVMAddIncoming(result, values.as_mut_ptr(), blocks.as_mut_ptr(), values.len() as u32);
+
+                // If all compared bytes were equal but sizes were not equal, the smaller vector is the lesser element.
+                let func = self.gen_cmp_fn(&Scalar(I64))?;
+                let left_size_ptr = LLVMBuildAlloca(builder, self.i64_type(), c_str!(""));
```
sppalkia: Same thing as above here. Since the code at the top is common, you could also just move the if statement to check if it supports memcmp to inside the vector match if you want to

weld/codegen/llvm2/cmp.rs
```diff
@@ -0,0 +1,323 @@
+//! Generates code to compare values using a comparison function.
+
+extern crate llvm_sys;
+
+use std::ffi::{CStr, CString};
+
+use ast::Type;
+use super::vector::VectorExt;
```
sppalkia: to support the change below, add

```rust
use super::vector;
```
above this line.

sppalkia:
Looks great! I'll post a follow up patch to this that updates the documentation etc. and merge this now.

Just for reference, the todos are:

* Shallow copy vectors before sorting
* Update `docs.md`
* Update Grizzly to use the new sort API.

PR #424, opened 2018-11-28
Sort Fixes

sppalkia:
Addresses a few issues with the newly added Sort functionality from #421:

* Adds MacOS compatibility
   - The `libc` `qsort_r` function takes arguments in a different order on Linux vs. FreeBSD, and also takes a comparator that takes arguments in a different order. This is fixed by checking the target OS when compiling.
* Clones a vector before sorting it.
* Fixes the constructor type checking in `constructors.rs`
* Updates `language.md` to reflect new sorting API
* Updates Grizzly code to use new sorting API (untested)


sppalkia:
@pratiksha I'm merging this since its causing tests on MacOS to fail, but you should take a look at it in case there are bugs in any of the comments.

pratiksha:
LGTM.

PR #425, opened 2018-12-07
Run cargo fmt

akshayknarayan:
This commit applies the cargo fmt style globally,
avoiding a mix of styles from various contributors.


sppalkia:
Thanks for the patch! I’m going to close this for now because we plan to upstream the active llvm-st branch over master soon (and the formatting changes will make this difficult), and we also have a branch working on upgrading the code there to respect the new lints and formatting tools.

PR #427, opened 2018-12-15
New Constructor API with associated functions

sppalkia:
Creates a more "Rusty" constructor API for expressions. This also changes the API to directly use the type checking logic in the `InferTypes` trait rather than effectively re-implementing type checking in the constructor like before.

This deprecates the `weld::ast::constructors` module (it will be removed eventually). Users should use the equivalent associated functions in the `NewExpr` trait instead. For example:

```rust
use ast::constructors;

let expr = constructors::binop_expr(BinOpKind::Add, left, right).unwrap();
```

Becomes

```rust
use ast::*;

let expr = Expr::new_bin_op(BinOpKind::Add, left, right).unwrap();
```

## Other Changes

* This patch also adds a `prelude` module to `ast`, to bring all structs/enums and their variants into scope at once.

* Moves some tests around

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,6 @@ install:
 - sudo rm -f `which llvm-config`
 - export WELD_HOME=`pwd`
 - PATH=$PATH:/home/travis/.cargo/bin
-- sudo -E apt-get -yq update &>> ~/apt-get-update.log
 - sudo -E apt-get --no-install-suggests --no-install-recommends -qy --allow-unauthenticated install gcc make llvm-6.0 llvm-6.0-dev clang-6.0 lib32z1-dev python2.7
   # - pip install virtualenv==15.1.0
   # - bash -e build_tools/travis/init_virtualenvs.sh 2.7 3.6
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,7 @@ install:
 - sudo rm -f `which llvm-config`
 - export WELD_HOME=`pwd`
 - PATH=$PATH:/home/travis/.cargo/bin
+- sudo -E apt-get -yq update
 - sudo -E apt-get --no-install-suggests --no-install-recommends -qy --allow-unauthenticated install gcc make llvm-6.0 llvm-6.0-dev clang-6.0 lib32z1-dev python2.7
   # - pip install virtualenv==15.1.0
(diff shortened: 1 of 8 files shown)


sppalkia:
@pratiksha you should take a look since it seems like you use the existing constructors API quite a bit.

sppalkia:
@pratiksha summarizing our conversation here:

* Seems strange to have `ExprKind` enumerated in multiple places, e.g., here and in `children` or `children_mut`. Would be nicer to have a systematic way of automatically doing this for new expression kinds.

* Discussed that procedural macros could be a way to do this. For the code here, we would manually write the code for `new_with_type` and `new`, which calls `infer_locally()`, and then generate code like this:

```rust
fn new_VARIANT( args ) -> Expr {
   Expr::new(ExprKind:new_VARIANT( args )
}
```

The `ExprKind::new_VARIANT` can be already be generated with the [`derive_new` crate](https://crates.io/crates/derive-new), but the wrapper over `Expr` will need to be custom (since the variants are of `ExprKind`, not `Expr`), and `derive_new` only looks at the variants of the annotated type).

The other challenge is mapping the `args` to `Box`'d values (or not) appropriately.

Another discussed item was just having a constructor for `ExprKind`, and asking the user to call `Expr::new` with it (which in turn fills out the type). This is not satisfying, however, because it leads to some fairly verbose code when building expressions. For example, to generate the expression `1+1`, this kind of API would need:

```rust
// new_literal is generated by the derive_new trait
let one = ExprKind::new_literal(LiteralKind(1)); 
let one = Expr::new(one);
// Note that the `Box` must be added here, since derive_new takes the exact fields
// in the generated function.
let added = ExprKind::new_binop(Add, Box::new(one.clone()), Box::new(one));
let added = Expr::new(added);
```

vs. the current API, which requires:

```rust
let one = Expr::new_literal(LiteralKind(1));
let added = Expr::new_binop(one, one.clone());
```

The difference isn't huge, but users will almost never want to construct an `ExprKind` on its own, so it seems strange to have constructors for it without the type. The second API is also arguably easier to wrap into an FFI.

Overall, it seems like the best thing to do to prevent writing per-expression code is to make _specialized_ procedural macros for generated this kind of boilerplate code. One thing to consider there is the readability etc., but that can be a later discussion.

@pratiksha feel free to add to this if I missed anything!

PR #429, opened 2019-02-03
Fix some memory leaks

Max-Meldrum:


weld/codegen/llvm2/jit.rs
```diff
@@ -219,12 +222,16 @@ unsafe fn verify_module(module: LLVMModuleRef) -> WeldResult<()> {
...
-    }
+    let result = {
+        if result_code != 0 {
+            let err = CStr::from_ptr(error_str).to_str().unwrap();
+            compile_err!("{}", format!("Module verification failed: {}", err))
+        } else {
+            Ok(())
+        }
+    };
+    let _ = CString::from_raw(error_str); // Free memory
```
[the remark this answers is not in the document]
sppalkia: You could also add a `libc::free(error_str)` on line 228, which should fix the leak.

weld/codegen/llvm2/jit.rs
```diff
@@ -243,21 +250,28 @@ unsafe fn optimize_module(module: LLVMModuleRef, conf: &ParsedConf) -> WeldResul
...
+        // TODO: Leaks memory
+        let lib_info = LLVMExtTargetLibraryInfo();
+        LLVMAddTargetLibraryInfo(lib_info, mpm);
         LLVMAddAnalysisPasses(target_machine, mpm);
         LLVMExtAddTargetPassConfig(target_machine, mpm);
         LLVMAddAnalysisPasses(target_machine, fpm);
     }
 
+    // Free memory
+    let _ = CString::from_raw(cpu_ptr);
```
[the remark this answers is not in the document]
sppalkia: You can replace them with:

```rust
libc::free(cpu_ptr);
libc::free(features_ptr);
```

Max-Meldrum:
Thanks for the comments. I've updated the code. I have left some TODOs where the calls are allocating memory using C++ new, although it is possible to libc::free them, it is usually not recommended.


leaks

weld/codegen/llvm2/jit.rs
```diff
@@ -69,24 +69,34 @@ impl CompiledModule {
...
+                if res == 1 {
+                    let err_str = CStr::from_ptr(err as *mut c_char)
+                        .to_string_lossy().into_owned();
+                    libc::free(err as *mut libc::c_void); // err is only allocated if res == 1
+                    compile_err!("Machine code generation failed with error {}", err_str)
+                } else {
+                    let start = LLVMGetBufferStart(output_buf);
+                    // TODO: output_buf Leaks. Is allocated with new and mismatches with libc::free
+                    let c_str = CStr::from_ptr(start as *mut c_char)
+                        .to_string_lossy().into_owned();
```
sppalkia: You can release `output_buf` with `LLVMDisposeMemoryBuffer ` after this line

weld/codegen/llvm2/jit.rs
```diff
@@ -69,24 +69,34 @@ impl CompiledModule {
...
-                let err = CStr::from_ptr(err as *mut c_char).to_str().unwrap();
-                return compile_err!("Machine code generation failed with error {}", err);
-            }
-            LLVMDisposeMessage(err);
-
-            let start = LLVMGetBufferStart(output_buf);
-            let c_str = CStr::from_ptr(start as *mut c_char);
-            Ok(c_str.to_string_lossy().into_owned())
+
+            let result = {
```
sppalkia: nit: get rid of the `let result = {` and just delete the if/else directly

weld/codegen/llvm2/jit.rs
```diff
@@ -69,24 +69,34 @@ impl CompiledModule {
...
+                    libc::free(err as *mut libc::c_void); // err is only allocated if res == 1
+                    compile_err!("Machine code generation failed with error {}", err_str)
+                } else {
+                    let start = LLVMGetBufferStart(output_buf);
+                    // TODO: output_buf Leaks. Is allocated with new and mismatches with libc::free
+                    let c_str = CStr::from_ptr(start as *mut c_char)
+                        .to_string_lossy().into_owned();
+                    Ok(c_str)
+                }
+            };
```
sppalkia: nit: delete the `}; result` (see above)

weld/codegen/llvm2/jit.rs
```diff
@@ -186,25 +196,31 @@ unsafe fn initialize() {
...
-    LLVMDisposeMessage(err);
-    Ok(LLVMCreateTargetMachine(target,
-                            PROCESS_TRIPLE.as_ptr(),
-                            HOST_CPU_NAME.as_ptr(),
-                            HOST_CPU_FEATURES.as_ptr(),
-                            LLVMCodeGenOptLevel::LLVMCodeGenLevelAggressive,
-                            LLVMRelocMode::LLVMRelocDefault,
-                            LLVMCodeModel::LLVMCodeModelDefault))
+    let ret = LLVMGetTargetFromTriple(PROCESS_TRIPLE.as_ptr(), &mut target, &mut err);
+    let result = {
```
sppalkia: nit: return inner if/else directly instead of assigning to result

weld/codegen/llvm2/jit.rs
```diff
@@ -219,12 +235,20 @@ unsafe fn verify_module(module: LLVMModuleRef) -> WeldResult<()> {
...
+            let err = CStr::from_ptr(error_str)
+                .to_string_lossy()
+                .into_owned();
+            compile_err!("{}", format!("Module verification failed: {}", err))
+        } else {
+            Ok(())
+        }
+    };
+    if !error_str.is_null() {
+        libc::free(error_str as *mut libc::c_void);
```
sppalkia: Move this above to before the error is returned (null check isn't needed since libc's free doesn't do anything if the pointer is null -- my mistake re: the comment earlier)

weld/codegen/llvm2/jit.rs
```diff
@@ -219,12 +235,20 @@ unsafe fn verify_module(module: LLVMModuleRef) -> WeldResult<()> {
     let result_code = LLVMVerifyModule(module,
                                        LLVMReturnStatusAction,
                                        &mut error_str);
-    if result_code != 0 {
-        let err = CStr::from_ptr(error_str).to_str().unwrap();
-        compile_err!("{}", format!("Module verification failed: {}", err))
-    } else {
-        Ok(())
+    let result = {
```
sppalkia: return if/else directly

weld/codegen/llvm2/jit.rs
sppalkia: Ah, you're right! No need to change this then (maybe get rid of the null check but not necessary)

Max-Meldrum:
Getting zero leaks now. Let me know if there is anything that should be modified

Cheers

sppalkia:
Looks great, thanks so much for fixing these!

PR #430, opened 2019-02-12
Add python3 support for grizzly

williamma12:
# What do these changes do

Adds python3 support for grizzly

### DIFF
--- a/python/grizzly/grizzly/dataframeweld.py
+++ b/python/grizzly/grizzly/dataframeweld.py
@@ -1,13 +1,13 @@
 import numpy as np
 import pandas as pd
 
-import grizzly_impl
-from lazy_op import LazyOpResult, to_weld_type
+from grizzly import grizzly_impl
+from .lazy_op import LazyOpResult, to_weld_type
 from weld.weldobject import *
-from utils import *
+from .utils import *
 
-from groupbyweld import GroupByWeld
-from seriesweld import SeriesWeld
+from .groupbyweld import GroupByWeld
+from .seriesweld import SeriesWeld
 
 class DataFrameWeld:
     """Summary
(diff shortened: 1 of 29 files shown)


python/grizzly/grizzly/dataframeweld.py
```diff
@@ -1,13 +1,13 @@
 import numpy as np
 import pandas as pd
 
-import grizzly_impl
-from lazy_op import LazyOpResult, to_weld_type
+from grizzly import grizzly_impl
```
williamma12: This does not work with python 2 for some reason. I am not really sure as to how to proceed with being able to support both python 2 and python 3 without putting it into a grizzly_impl class.

sppalkia:
Hey Will,

This looks pretty good, I don't have any specific comments. Just to clarify, this will break things with Python2, right? It may be worth adding a comment about this in the README since it will likely break all the existing [benchmarks](https://github.com/weld-project/weld-benchmarks) and [examples](https://github.com/weld-project/weld/tree/master/examples/python/grizzly).

In fact, for that reason, it may make more sense to keep this as a separate branch altogether, where the Weld compiler part is kept in sync with `llvm-st` and the Python3 dev happens there. We can merge it back in after updating the benchmarks etc.

sppalkia:
The sort/groupBy may have broken due to an earlier change (we haven't been testing Grizzly as actively as the main compiler) -- I can post a patch for that in a different PR. Thanks for pointing it out!

python/grizzly/grizzly/dataframeweld.py
radujica: Have you tried adding grizzly_impl import in __init__ and then here doing `from . import grizzly_impl`?

I'd still recommend using `from .grizzly_impl import <functions>` and only relative imports within the module.

williamma12: ah thanks! the `from . import grizzly_impl` worked

sppalkia:
From @williamma12:

> before we can merge python3 branch into llvm-st, we’ll have to fix the circular dependencies in grizzly and fix the fact that python3 wont find `SUPPORTED_DTYPES` after doing `from weldnumpy import *`

PR #431, opened 2019-02-13
Add support for Type name aliasing

sppalkia:
Adds support for aliases for types ("typename aliases"). Aliases must currently be listed before macros before the expression representing the Weld program:

```
type int = i32;
type pair = {i32,i32};

# Define macros here.

|v: int, p: pair| v + p.$0 + p.$1
```

Typename aliases can be used within other typename aliases (as long as they are defined before):

```
type int = i32;
type pair = {int,int};

|v: int, p: pair| v + p.$0 + p.$1
```

Macros can also use typename aliases:

```
type int = i32;

macro addOne(a) = (
  cudf[addOneUdf,int](a)
);

# Expands to |v: i64| cudf[addOneUdf,i64](v)
|v: int| addOne(v)
```

Currently, typename aliases are treated similarly to macros: when a program is compiled, type typenames are replaced with their true types, and the naming information is lost. This means that dumping code, logging messages, etc. will currently show the actual type rather than the typename. We hope to propagate type name information through the compiler soon.


docs/language.md
```diff
@@ -272,10 +275,10 @@ In particular, Weld only requires types for the top-level function arguments, an
       )
 ```
 
-# Sugar Operations
+# Sugar Operations and Macros
 
-To make programs easier to write, the Weld implementation also supports some "sugar" operations that translate into `for`s and builders.
-These are currently represented as *macros*, which are substitution rules whose definitions are not handled by the optimizer.
+To make programs easier to write, Weld also supports some "sugar" operations that translate into `for`s and builders.
+Most of are currently represented as *macros*, which are substitution rules whose definitions are not handled by the optimizer.
```
pratiksha: "Most of these"

docs/language.md
```diff
@@ -287,7 +290,7 @@ Signature | Notes
 `zip(v1: vec[T1], v2: vec[T2], ...): vec[{T1, T2, ...}]` | Only allowed in the `vec` argument of the `for` loop.
 `compare(x: T, y: T)` | Implements a default comparator for `sort`. Expands to `if(x > y, 1, if(x < y, -1, 0))`.
 
-All of these operations can straightforwardly be translated into `for` expressions.
+With the exception of `zip` (which is a special syntactic sugar that isn't expanded from a macro), all of these operations can be translated into `for` expressions.
```
pratiksha: it is a little confusing to have `zip` here because the consequences are very different from macros. maybe consider separating it into its own subsection

docs/language.md
```diff
@@ -302,6 +305,61 @@ macro filter(data, func) = (
...
+macro doubleAdd(a, b) = (
+  a + a + b + b
+);
+
+|v1: i32, v2: i32|
+  # Expands to v1 + v1 + v2 + v2
+  doubleAdd(v1 ,v2)
+```
+
+Macros are _hygenic_, so variable names defined within a macro will never clash (and by extension cannot be accessed from outside the macro expansion).
```
pratiksha: hygienic

docs/language.md
```diff
@@ -302,6 +305,61 @@ macro filter(data, func) = (
...
+Since macros do not require types in their parameters, they are a useful way to leverage type inference and write templatized Weld code.
```
pratiksha: Do top-level Weld lambdas have to have typed parameters? If so this seems to need a clarification (i.e. the entire program can't just be a macro call, even though a macro is technically a complete Weld program).

weld/syntax/macro_processor.rs
```diff
@@ -136,67 +212,141 @@ fn update_defined_ids(expr: &mut Expr, sym_gen: &mut SymbolGenerator) {
...
+#[test]
+fn redefinition_with_type_alias_in_type_alias() {
+    // Types are defined in order: redefining a type shouldn't redefine
+    // types that occur before it.
+    let aliases = parse_type_aliases("type int = i32;
+                                     type pair = {int,int};
+                                     type int = i64;").unwrap();
+    let mut expr = parse_expr("|a: pair| a").unwrap();
+    process_type_aliases(&mut expr, aliases).unwrap();
+    assert_eq!(print_expr_without_indent(&expr).as_str(), "|a:{i32,i32}|a");
```
pratiksha: suggest adding:
```
  let mut expr = parse_expr("|b: int| b").unwrap();
  process_type_aliases(&mut expr, aliases).unwrap();
  assert_eq!(print_expr_without_indent(&expr).as_str(), "|b:i64|b");
```
if that is indeed the intended behavior

weld/syntax/program.rs
```diff
@@ -1,17 +1,27 @@
...
+use ast::{Expr, Symbol, Type};
 
 #[derive(Clone, Debug, PartialEq)]
 pub struct Program {
     pub macros: Vec<Macro>,
+    pub type_aliases: Vec<TypeAlias>,
     /// Program body -- this will likely be a Lambda, but not always.
     pub body: Expr,
 }
+/// A type alias we will substitate at compile time.
```
pratiksha: substitute

weld/syntax/parser.rs
```diff
@@ -1212,6 +1247,13 @@ impl<'t> Parser<'t> {
         let mut annotations = Annotations::new();
         try!(self.parse_annotations(&mut annotations));
 
+        if !self.peek().signals_type() {
+            let name = self.symbol()?.to_string();
+            // NOTE: The type field is not used right now, but may be later if we decide
+            // to keep type alias information past macro substitution.
+            return Ok(Alias(name, Box::new(Unknown)));
```
pratiksha: One question here: it seems that only the last redeclaration of the same name should be preserved in the AST, otherwise there might be issues resolving the type (depending on whether ordering information is preserved / considered during type inference)?

weld/syntax/macro_processor.rs
sppalkia: I think it makes more sense for the definitions to be processed in order (there's actually some extra logic I added to make sure that was the case) -- otherwise it becomes confusing since later definitions modify earlier ones.

weld/syntax/parser.rs
sppalkia: Each name should have a single type by the type the expression is processed (namely, the last definition for the name), so there shouldn't be conflicts. The definitions themselves won't ever have aliases since they're processed iteratively. E.g., if you have:

```
type int = i32;
type pair = {int,int};
type pairpair = {pair,pair};
```

When `pairpair` is processed, `pair` is already processed to be `{i32,i32}`.

docs/language.md
sppalkia: Good point, I'll just delete this comment.

docs/language.md
sppalkia: Hmm yeah, they're already discussed in the for loop section, so I just deleted the reference to this here.

pratiksha:
lgtm!

PR #432, opened 2019-02-13
Add assertion operator

sppalkia:
Adds an `assert` operator. The assert expression evaluates to `true` if the assertion condition is `true`, and terminates the program with an `AssertionError` otherwise.

Example Usage:

```
|v: i32| assert(v == 0)
```


pratiksha:
Looks great!

sppalkia:
Thanks for looking at it!

PR #437, opened 2019-02-28
LLVM External Error fix for Ubuntu

sppalkia:
This patch registers functions using the globals API in LLVM to fix the "external error" issue (#394).

@Max-Meldrum could you confirm that this works on your machine for your code? I tried it with the example on 16.04 on EC2 and it works there, but just to double check.


Max-Meldrum:
@sppalkia Tested it and it is working (18.04). Thanks for having a look at this!

sppalkia:
@Max-Meldrum thanks for testing it!

PR #447, opened 2019-05-12
Change serialize/deserialize to u8

### LINKED ISSUE #446: Weld Serialize/Deserialize
I am using Weld's `serialize(data)` and  `deserialize[T](data)` in a project. As most other libraries handle bytes as `Vec<u8>`, I am currently having to cast between i8/u8 while going from Weld to other libraries (e.g., RocksDB, Protobuf) and vice versa. Is there a motivation behind the choice of `Vec<i8>`?

I could look into creating a PR for this. Or if it would be breaking other projects relying on the current implementation, I could work on a fork instead.

Cheers

Max-Meldrum:
I think that should be all necessary changes. But let me know if you see any missing changes.

closes #446 

All serialization examples in the cpp run accordingly. Also, the weld (rust) serialization tests pass as well.

### DIFF
--- a/docs/language.md
+++ b/docs/language.md
@@ -130,8 +130,8 @@ refer to types.
   value until the boolean it returns is `false`, and then return the `T` field in
   its output as the final value of the expression.
 * `cudf[name,ty](args)` to call arbitrary C-style functions (see a discussion of UDFs [below](#user-defined-functions)).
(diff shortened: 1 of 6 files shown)


sppalkia:
Looks good, thanks for adding this!

PR #452, opened 2019-06-08
Fix instructions and dependencies on macos

winding-lines:
Use `brew --prefix` instead of hardcoded path, it does not work on my system.
Fix the flags for the c++ compiler

Ran `cargo build --release` on macos mojave and ubuntu 18.04


sppalkia:
Thanks for the fixes! It looks good to me -- is it ready to merge?

winding-lines:
Ready!

PR #453, opened 2019-06-13
add namespace std for sort function in llvmext.cpp

harumichi:
because compiling in some environment is failed due to name ambiguous error.


harumichi:
llvm: 8.0.0

PR #479, opened 2019-09-25
Generate core Weld API Python bindings with PyO3

sppalkia:
This patch generates Python 3 compatible bindings for the main Weld API using PyO3. This makes the bindings slightly more robust, and also makes building them easier since PyO3 contains tools to automate setup (e.g., without having to set an environment variable).

One issue is that Weld currently uses stable Rust, while PyO3 requires the specialization features from nightly Rust. This module may thus be distributed separately from the main Weld workspace, or distributed as a different repository altogether.

Addresses #485


sppalkia:
Note that this package (located under `weld-python` is a work in progress! It's being merged to make working on other related issues (see [milestone](https://github.com/weld-project/weld/milestone/1)).

PR #504, opened 2020-03-12
Add function decorator to make Python Weld codegen easier

sppalkia:
This patch adds a new function decorator `weldfunc` that converts a function that returns strings of Weld code into functions that return `WeldLazy`. The function automatically detects inputs that are themselves `WeldLazy`, and adds them as dependencies to the return value of the wrapped function. For example, say we have:

```python
def exp(a, b):
   return "exp({}, {})".format(a, b)
```

This function (or its caller, as is done right now in Grizzly) would ordinarily need to check whether `a` or `b` are `WeldLazy` objects, extract the name, construct the dependency list, etc.:

```python
val1 = 5.0
val2 = ... # create a WeldLazy
code = exp(val1, val2)
dependencies = [val2]
result = WeldLazy(code, dependencies, output_type, decoder)
```

Instead of doing this manually, decorating `exp` with `weldfunc` will construct the dependency list automatically and return a `WeldLazy`:

```python
@weldfunc # With this, exp can assume all inputs are strings now
def exp(a, b):
   return "exp({}, {})".format(a, b)

# Code to create result as above
val1 = 5.0
val2 = # create a WeldLazy
result = exp(val1, val2)(output_type, decoder)
```

Note that the return value of `exp(val1, val2)` is itself a callable, so you can still specify a custom decoder/output type (eventually, the hope is that the output type can be inferred in all cases with a nicer API, but since we have strings for now, this will do). `result` will already assign `val2` as a dependency above. Users can still get the raw code constructed by `exp` with `exp(val1, val2).code` (this is mostly just useful for tests).

### How is this patch tested?

* New test for `weldfunc` to check dependencies of returned value
* Existing E2E tests
* New doctests.


weld-python/weld/lazy.py
```diff
@@ -301,3 +302,83 @@ def identity(phys_value, decoder):
...
+class WeldFunc(object):
+    """
+    The result of a function annotated with `welfunc`.
+
+    This object is effectively a function with a field `code`, which extracts the raw code
+    constructed by the annotated function without constructing a `WeldLazy`.
+
+    """
+
+    __slots__ = ["_code", "_dependencies"]
```
deepakn94: Why do you need this?

weld-python/tests/weld/lazy/test_weldfunc.py
```diff
@@ -0,0 +1,51 @@
+
+from weld import *
+from weld.lazy import WeldLazy, weldfunc
+from weld.types import *
+
+def test_weldfunc_simple():
```
deepakn94: Can you write a test with 2 or more dependencies as well? And a longer chain with a single dependency each?

weld-python/weld/lazy.py
```diff
@@ -301,3 +302,83 @@ def identity(phys_value, decoder):
...
+    >>> v1 in v2.children
+    True
+
+    """
+    @functools.wraps(func)
+    def weldfunc_wrapper(*args, **kwargs):
+        new_args = list()
+        new_kwargs = dict()
+        dependencies = []
+        for arg in args:
```
deepakn94: Comment here saying you're peeling all `WeldLazy` arguments, and updating dependencies as needed.

weld-python/weld/lazy.py
```diff
@@ -301,3 +302,83 @@ def identity(phys_value, decoder):
     Creates an identity `WeldLazy` from a `PhysicalValue` and decoder.
     """
     return WeldLazy(str(phys_value.id), [phys_value], phys_value.output_type, decoder)
+
+class WeldFunc(object):
+    """
+    The result of a function annotated with `welfunc`.
```
deepakn94: Typo: `weldfunc`

weld-python/weld/lazy.py
```diff
@@ -301,3 +302,83 @@ def identity(phys_value, decoder):
...
+
+    This object is effectively a function with a field `code`, which extracts the raw code
+    constructed by the annotated function without constructing a `WeldLazy`.
```
deepakn94: Can you add a comment that `WeldFunc` object shouldn't be consumed by anything? Rather, it's expected that the `__callable__` function is called on this object to get a `WeldLazy` object that can be used downstream.

weld-python/weld/lazy.py
```diff
@@ -301,3 +302,83 @@ def identity(phys_value, decoder):
...
+    """
+
+    __slots__ = ["_code", "_dependencies"]
+
+    def __init__(self, code, dependencies):
+        self._code = code
+        self._dependencies = dependencies
+
+    @property
+    def code(self):
```
deepakn94: You don't want to add a property for the dependencies themselves? Might be useful in tests?

weld-python/weld/lazy.py
```diff
@@ -301,3 +302,83 @@ def identity(phys_value, decoder):
...
+    >>> v1.expression
+    'sqrt(1)'
+    >>> v2 = sqrt(v1)(F64(), None) # v1 becomes a dependency of v2
+    >>> v2.expression == "sqrt({})".format(v1.id)
+    True
+    >>> v1 in v2.children
+    True
+
+    """
+    @functools.wraps(func)
```
deepakn94: Comment on what this does? Seems to ensure that the wrapped function's name, etc. is not overwritten by the wrapper?

weld-python/weld/lazy.py
sppalkia: This represents the object as a tuple instead of a dict, so its slightly more efficient to construct the object (it also prevents adding other attributes unless the `__slots__` definition is updated). Using it here since we don't want to make the creation of this wrapper object expensive.

weld-python/weld/lazy.py
sppalkia: Hmm, can you clarify? The API for this doesn't prevent a caller from accessing code or doing something else with this (though it's not very useful for anything other than calling)

weld-python/weld/lazy.py
sppalkia: I think its okay for now, since that field is not used -- no point in adding dead code.

weld-python/weld/lazy.py
deepakn94: Feel free to ignore this. Just wanted a clarifying comment saying dependencies are tracked using `WeldLazy`, and you need to call the `__callable__` function to get this.

weld-python/weld/lazy.py
deepakn94: Interesting

deepakn94:
Cool, feel free to merge!

sppalkia:
@deepakn94 thanks for the review!

Issue #507, opened 2020-03-19
"simplify_assignments" creates infinite loop with iterate


sppalkia:
e.g., the following code will loop infinitely due to an incorrect variable deletion in the SIR:

```
|e: vec[i8]|
let lenString = len(e);
iterate(lenString - 1L, |p| {p - 1L, p > 7L})
```

PR #508, opened 2020-03-19
Add string functions

sppalkia:
This adds support for processing Series composed of string data.

## TODO

- [x] Add tests
- [x] Add remaining functions
- [ ] Fix assignment simplification patch (#507)


weld-python/weld/grizzly/core/strings.py
```diff
@@ -0,0 +1,263 @@
...
+
+    def get(self, index):
+        """
+        Get the character at index 'i' from each string. If 'index' is greater than
+        the string length, this returns an empty string. If 'index' is less than 0,
+        this wraps around, using Python's indexing behavior.
+
+        Examples
+        --------
+        >>> from weld.grizzly import GrizzlySeries
```
sppalkia: @deepakn94 any idea how to fix this? Seems like there should be a way to add a common prelude to doctests, but didn't find how to do it...

sppalkia: See weld-python/weld/conftest.py for how to do this for future reference (thanks @paroma!)

deepakn94: Interesting -- good to know for the future!

PR #513, opened 2020-09-09
add -fPIE to support clang 10.0+

harbby:


harbby:
add clang 10.0+ support

